PR #9, opened 2020-11-25
Add CNF benefits as part of the WG scope

electrocucaracha:
This group should be consider the importance of implementing Cloud-Native architecture on Network Functions. The definition provides an answer to the **what** question but listing some benefits can help to answer the _why_ question.

### DIFF
--- a/charter.md
+++ b/charter.md
@@ -18,6 +18,7 @@ Cloud Native Network Function Working Group’s mission is to increase interoper
 
 ## In Scope
 - Definition of Cloud Native Network Function (CNF)
+  - Provide a list of the cloud native benefits a best practice provides
 - Cloud native conformance test requirements for CNFs
   - Including dataplane CNFs
 - Process around certifying CNF conformance


taylor (approved):
needs more thought...

taylor (changes requested):
@electrocucaracha this sounds like a good item for "Why" on a best practice.  Maybe a benefit could be written about in a white paper which is in scope but not required.  

Basically the Why is something we want to relate things to but this group is not trying to convince everyone of the general benefits of cloud native.  The Telecom User Group has that as a goal -- eg. whitepapers.  Other groups are also doing this. The Cloud Native Networking Principles papers are related to this.

I suggest moving this to an item in the proposal template or somewhere encouraging references for best practices which are proposed and discussed.

electrocucaracha:
> @electrocucaracha this sounds like a good item for "Why" on a best practice. Maybe a benefit could be written about in a white paper which is in scope but not required.
> 
> Basically the Why is something we want to relate things to but this group is not trying to convince everyone of the general benefits of cloud native. The Telecom User Group has that as a goal -- eg. whitepapers. Other groups are also doing this. The Cloud Native Networking Principles papers are related to this.
> 
> I suggest moving this to an item in the proposal template or somewhere encouraging references for best practices which are proposed and discussed.

@taylor if I understand correctly what you're saying that we have [this principles](https://github.com/cncf/cnf-wg/blob/master/charter.md#mission-statement) but new principles have to justify their addition, for example something like to use `WebAssembly` instead of containers may improve performance or security but affect portability so the proposal has to justify that right?

jeffsaelens:
I think mapping benefits is a good idea, but I agree with @taylor that we probably need to put on our TUG hats and not let those efforts die on the vine. For good or ill, I think the predetermined choices at the moment are K8s and a container runtime. Purists like myself wish we could crawl down the uni-kernel and FaaS/serverless rabbit holes, but that just isn't practical from a timeline perspective at the moment.

A proposal like moving to web assembly happens at an echelon or two above our group I think. We collaborate on a perspective in the TUG and try and shift the paradigm at a larger scale. I don't see any harm though in potentially listing the cloud native benefits a best practice might bring in addition to the requirements/principles/pillars/<whatever we are calling these golden rules> that are being met. If anything, it might provide more insight in to what the cloud native approach is bringing to the table in addition to the satisfied requisites.

charter.md
```diff
@@ -18,6 +18,7 @@ Cloud Native Network Function Working Group’s mission is to increase interoper
 
 ## In Scope
 - Definition of Cloud Native Network Function (CNF)
+  - Provide a list of benefits of migrating to/implementing a Cloud-Native architecture
```
jeffsaelens: maybe change the scope of this to say list the cloud native benefits a best practice provides? I think for big picture items we could reference a lot of existing docs that OpenInfra, Linux Foundation and the big public cloud players.

jeffsaelens (changes requested):
Let me know your thoughts. Can we reference existing docs on the benefits of going cloud native in the general context and instead try being more specific in this WG's efforts to map which benefit a practice may bring about?

charter.md
electrocucaracha: You mean like this?
```suggestion
  - Provide a list the cloud native benefits a best practice provides
```

jeffsaelens: I dig it. just add an "of" in front of "the cloud"

xmulligan: I agree. If you change that, We can merge that

PR #11, opened 2020-11-30
Adding the Deutsche Telekom as interested party

vukg:
As representative of Deutsche Telekom Technik's Kubernetes Engine team, I would like to rise our intention to take active part in this working group.


xmulligan:
Happy to have you :)

PR #12, opened 2020-11-30
Add Cloud-Native "motivation" to group goals

rannyh:
Based on the call on Monday, November 30th, it sounds like we first need to agree on the motivation for adopting cloud-nativeness, before going into the nitty-gritty details of how to implement a cloud-native CNF.

### DIFF
--- a/charter.md
+++ b/charter.md
@@ -6,7 +6,7 @@ The goal of the Cloud Native Network Function Working Group (CNF WG)  is to aid 
 The CNF WG operates under the aegis of CNCF. The charter of the working group is to define the process around certifying the cloud nativeness of networking applications, aka CNFs. We collaborate with the [CNF test suite project](https://github.com/cncf/cnf-conformance/blob/master/README-testsuite.md) who works on the mechanics of the conformance tests.
 
 The goals for the group are:
-- To build concensus around the motivation for using Cloud-Native architecture and DevOps methodologies for telecommunications.
+- To build consensus around the benefits of using cloud native architecture methodologies for telecommunications.
 - To create a software conformance program that any application or network function implementation can use to demonstrate that they are conformant and interoperable with cloud native principles.
 
 
--- a/charter.md
+++ b/charter.md
@@ -5,7 +5,10 @@ The goal of the Cloud Native Network Function Working Group (CNF WG)  is to aid 
 
 The CNF WG operates under the aegis of CNCF. The charter of the working group is to define the process around certifying the cloud nativeness of networking applications, aka CNFs. We collaborate with the [CNF test suite project](https://github.com/cncf/cnf-conformance/blob/master/README-testsuite.md) who works on the mechanics of the conformance tests.
 
-The goal for the group is to create a software conformance program that any application or network function implementation can use to demonstrate that they are conformant and interoperable with cloud native principles.
+The goals for the group are:
+- To build concensus around the motivation for using Cloud-Native architecture and DevOps methodologies for telecommunications.
+- To create a software conformance program that any application or network function implementation can use to demonstrate that they are conformant and interoperable with cloud native principles.
+
 
 ## Mission Statement
 Cloud Native Network Function Working Group’s mission is to increase interoperability and standardization of cloud native workloads. It is committed to the following (aspirational) design ideals:
--- a/charter.md
+++ b/charter.md
@@ -7,7 +7,7 @@ The CNF WG operates under the aegis of CNCF. The charter of the working group is
 
 The goals for the group are:
 - To build consensus around the benefits of using cloud native architecture methodologies for telecommunications.
-- To create a software conformance program that any application or network function implementation can use to demonstrate that they are conformant and interoperable with cloud native principles.
+- To create a program that identifies best practices for CNFs running on K8s which CNF Developers and operators alike may use to demonstrate how well their software and its operations adhere to cloud native principles.
 
 
 ## Mission Statement


charter.md
```diff
@@ -5,7 +5,10 @@ The goal of the Cloud Native Network Function Working Group (CNF WG)  is to aid
 
 The CNF WG operates under the aegis of CNCF. The charter of the working group is to define the process around certifying the cloud nativeness of networking applications, aka CNFs. We collaborate with the [CNF test suite project](https://github.com/cncf/cnf-conformance/blob/master/README-testsuite.md) who works on the mechanics of the conformance tests.
 
-The goal for the group is to create a software conformance program that any application or network function implementation can use to demonstrate that they are conformant and interoperable with cloud native principles.
+The goals for the group are:
+- To build concensus around the motivation for using Cloud-Native architecture and DevOps methodologies for telecommunications.
```
jeffsaelens: I feel like this ship has sailed. Motivated or not, this technology shift is coming. Vendors are not going to provide offers that don't come packaged this way in the future. I think showing how cloud native practices map to service provider business outcomes would help. Up time is our numero uno. Pointing to self-healing services is an example on how this shift meets the architectural constraints of CSPs.

charter.md
```diff
@@ -5,7 +5,10 @@ The goal of the Cloud Native Network Function Working Group (CNF WG)  is to aid
...
+- To create a software conformance program that any application or network function implementation can use to demonstrate that they are conformant and interoperable with cloud native principles.
```
jeffsaelens: Sounded like we are dropping conformance for now and moving to best practices. Maybe we shift the focus? "To provide a repository of best practices for CNF developers and operators alike, allowing them to demonstrate that the software and the operation of said function strives to adhere to cloud native principles."?

jeffsaelens (changes requested):
I think we should revisit these after Monday's call.

charter.md
taylor: @jeffsaelens good catch. We do want to update the language to focus on best practices.


How about something like this:
- To create a program which identifies best practices for CNFs running on K8s which CNF Developers and operators alike may use to demonstrate how well their software and its operation adhere to cloud native principles.
- To determine which practices will allow telco applications to most efficiently utilize K8s to lower the operational risk and burdens of service providers and telco vendors

charter.md
rannyh: Well, not so fast. I believe it is still a good drill to start with the motivation. You mentioned modern packaging and uptime. Some network functions (as you probably witnessed yourself) are just ported from VM's to Docker images, and continue using their 1+1 (on n+m) redundancy schemes, claiming they are "cloud-native". We want to avoid exactly that, and point to the tenets of cloud-native that provide the proper functionality. Also, I heard different cloud-native motivations from some of the other operators, like expecting vendors to provide smaller units of functionality, so they can integrate their own network functions using best of breed components. Not saying I subscribe to this approach, but  it exists anyway. Starting from agreeing on the motivation will, IMHO, keep this effort focused.

taylor: @rannyh while I think it will be good for the best practices which are discussed to include the motivations and benefits for the telecommunication industry I think it should not be stated as a primary goal.  Please see the [TUG whitepaper](https://github.com/cncf/telecom-user-group/blob/master/whitepaper/CNCF%20_TUG_Cloud%20Native%20Thinking%20for%20Telecommunications.pdf), or the multitude of books, blog posts, and videos for motivations in using cloud native and devops methodologies.

That said, [in scope for the CNF WG](https://github.com/cncf/cnf-wg/blob/master/charter.md#in-scope), is "Publishing metrics/white papers". So if a paper relating the cloud native benefits in telco terms for motivation would be beneficial that can be done.

It is not the goal / focus of this group though.

The primary goal is to help folks who want to obtain the benefits of cloud native principles and methodologies in gaining them.

taylor (changes requested):
Motivation could be added to in-scope and/or to the proposal template.  It is not a primary goal for the working group.

charter.md
jeffsaelens: @rannyh I don't think we are saying different things. Specifically, I am saying that I think the "do I go cloud-native?" at the macro level is sort of behind us. Individual motivations as to whether we go down path "x" or "y" based on the cloudiness of something is a different matter. That in turn ties to those business drivers. I feel like none of us would be here in these numbers if there wasn't already strong motivation to migrate towards a more cloud-native approach in general.

taylor: @rannyh I agree that having some context on the motivations will help.  I am saying it is secondary to more primary goals such as does a best practice provide enough context to communicate its benefits to creators, vendors, and operators.

We can have papers in the repo that help with context, add links and short text into the README though.

rannyh: @Taylor and @jeffsaelens - I am OK with not making "defining motivation" part of the goals. I never meant it to be an answer to "why go cloud-native?" but rather some sort of consensus on "which of the cloud-native tenets are relevant to CSPs?". I do see a situation where we may have a hard time agreeing on some "best practice" down the road. Having an agreed set of motivations could help resolve such debates. Maybe it is already covered well enough in the TUG Whitepaper and we should not repeat it here.

taylor: @rannyh I think that if someone proposes a best practice for a telco audience (dev, ops, etc) then the proposal should have the motivation and benefits listed. There should be references for why it's a best practice. Its part of the process and templates now.  If a best practice is not relevant to CSPs then that should come out in the discussion of a proposal.

Then in the meetings we can discuss proposals that have been put forth or bring up new ones.  More time to discuss how a best practice provides benefits to CSPs and any issues and alternatives.

jeffsaelens: I've been calling these requirements in my head. Does best practice "A" align with a CSP (or any network operator) requirement. This could have been misreading on my part, but it seemed a deep and bottomless pit to fall down in to if we were debating if we should be doing cloud-native-y things in the first place.

taylor: > Does best practice "A" align with a CSP (or any network operator) requirement. 

And what requirement does this best practice help fulfill.

electrocucaracha: > I feel like this ship has sailed. Motivated or not, this technology shift is coming. Vendors are not going to provide offers that don't come packaged this way in the future. I think showing how cloud native practices map to service provider business outcomes would help. Up time is our numero uno. Pointing to self-healing services is an example on how this shift meets the architectural constraints of CSPs.

IMHO, `Up time` is hard to define here, containers are ephemeral so their up time is pointless as long as they can recover quickly, in other words, fail-fast but boot instantly. AFAIK, the main driver to use containers was major density on servers. I like the @taylor 's idea about accepting new proposals but we have define the initial requirements.

jeffsaelens: when I think of uptime, I'm usually specifically referring to, "is the service reachable?". We can hide how the sausage is made from the consumer of the service in most instances. "Does my phone have 4g? yep! we're good here."

charter.md
xmulligan:
```suggestion
- To create a program that identifies best practices for CNFs running on K8s which CNF Developers and operators alike may use to demonstrate how well their software and its operations adhere to cloud native principles.
```

charter.md
xmulligan:
```suggestion
- To build consensus around the benefits of using cloud native architecture methodologies for telecommunications.
```

xmulligan:
@rannyh I've added some suggestions, let me know what you think

charter.md
```diff
@@ -5,7 +5,10 @@ The goal of the Cloud Native Network Function Working Group (CNF WG)  is to aid
 
 The CNF WG operates under the aegis of CNCF. The charter of the working group is to define the process around certifying the cloud nativeness of networking applications, aka CNFs. We collaborate with the [CNF test suite project](https://github.com/cncf/cnf-conformance/blob/master/README-testsuite.md) who works on the mechanics of the conformance tests.
 
-The goal for the group is to create a software conformance program that any application or network function implementation can use to demonstrate that they are conformant and interoperable with cloud native principles.
+The goals for the group are:
+- To build consensus around the benefits of using cloud native architecture methodologies for telecommunications.
```
xmulligan, line 9:
```suggestion
- To build consensus around the benefits of using cloud native architecture methodologies for telecommunications through user stories and use cases.
```

taylor: I think adding specific work items/deliverables like this would be good under the charters [In Scope](https://github.com/cncf/cnf-wg/blob/f6bd97dbcf145d2b1ad6ebf2e27da3d5538f2c84/charter.md#in-scope) section. The high-level goal could stay general rather than specifying the method for achieving it.

charter.md
```diff
@@ -5,7 +5,10 @@ The goal of the Cloud Native Network Function Working Group (CNF WG)  is to aid
 
 The CNF WG operates under the aegis of CNCF. The charter of the working group is to define the process around certifying the cloud nativeness of networking applications, aka CNFs. We collaborate with the [CNF test suite project](https://github.com/cncf/cnf-conformance/blob/master/README-testsuite.md) who works on the mechanics of the conformance tests.
 
-The goal for the group is to create a software conformance program that any application or network function implementation can use to demonstrate that they are conformant and interoperable with cloud native principles.
```
taylor, line 8: Note: This was already deleted. This PR is not against current charter.md

charter.md
taylor, line 9: This was a goal for the TUG and other efforts.  It seems like a repeat in the CNF WG. It is a sub-goal of every best practice (eg. what are the benefits of this best practice).

The primary audience for the CNF WG is for people that want to *gain* benefits and are looking for practices, methodologies, etc which will provide them.


If the goal is going to be listed it should be related to the best practices goal and come after it.  Something like:

```suggestion
- To highlight the benefits of using cloud native best practices for telecommunications
```

taylor (commented):
I suggest a new PR be created if desired based on the most recent version of the charter.md and this PR be closed.

taylor:
> I suggest a new PR be created if desired based on the most recent version of the charter.md and this PR be closed.

I've created a new PR with partial updates (not covering all items here), but making it a smaller, more concise focus for the goals https://github.com/cncf/cnf-wg/pull/47

xmulligan:
@rannyh are you ok if we close this request and move the discussion over to #47 ?

xmulligan:
You were just a hair faster than me :D

PR #15, opened 2020-12-02
updated grammar and redundancy

agentpoyo:


taylor (approved):
Thanks @agentpoyo.  Its better than what was there.  Maybe selected from vs following would be better.

PR #29, opened 2020-12-09
Provide context for the type of CNF the best practice applies to in 0001-cnf-conformance-definition-proposal-process.md

jeffsaelens:
First pass here before looking at the template itself. I really think that a best practice should lay out to type of workload or container it applies to. OSS/BSS, NF and infra containers may have best practices applied to them in unequal measure depending on their desired outcome. A container specifically designed to provision infra will need privilege for example, while a container hosting BGP would not. 

They alternative to this approach is for the definition of a CNF to be formalized (producing this is listed as a group outcome in our charter) and have its scope limited. From there we clarify that the best practices specifically apply to these "types" of workloads/containers.

### DIFF
--- a/ccdps/0001-cnf-conformance-definition-proposal-process.md
+++ b/ccdps/0001-cnf-conformance-definition-proposal-process.md
@@ -117,7 +117,7 @@ In addition to the challenge of managing issues over time, searching for text wi
 
 ## **Workload Context**
 
-Context should be provided for the type of workload/container this best practice applies to. The context, of an individual best practice, should illustrate the specific type of workload most likey to take advantage of the benefits listed. In contrast, it should provide a warning if the practice would be a determent to a divergent type of workload.
+Context should be provided which includes the type of CNF, workload, and CNF component (eg. Pod, container, Operator) this best practice applies to. The context, of an individual best practice, should illustrate the specific type of workload most likely to take advantage of the benefits listed. When applicable, it should also provide a warning if the practice would be a determent to a divergent type of workload.
 
 
 ## **Test Plan**
--- a/ccdps/0001-cnf-conformance-definition-proposal-process.md
+++ b/ccdps/0001-cnf-conformance-definition-proposal-process.md
@@ -12,6 +12,7 @@
   - [Notes/Constraints/Caveats (Optional)](#notesconstraintscaveats-optional)
   - [References](#references)
   - [Alternatives (Optional)](#drawbacksalternatives)
+- [Workload Context](#workload-context)
 - [Test Plan](#test-plan)
 - [Scoring](#scoring)
 - [Implementation History](#implementation-history)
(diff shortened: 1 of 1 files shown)


ccdps/0001-cnf-conformance-definition-proposal-process.md
```diff
@@ -113,10 +115,16 @@ The use of GitHub issues when proposing changes does not provide the CNF WG good
...
 
 
+## **Workload Context**
+
+Context should be provided for the type of workload/container this best practice applies to. The context, of an individual best practice, should illustrate the specific type of workload most likey to take advantage of the benefits listed. In contrast, it should provide a warning if the practice would be a determent to a divergent type of workload.
+
+
```
taylor: @jeffsaelens can you provide some examples of different types?

ccdps/0001-cnf-conformance-definition-proposal-process.md
```diff
@@ -113,10 +115,16 @@ The use of GitHub issues when proposing changes does not provide the CNF WG good
...
 
 
+## **Workload Context**
+
+Context should be provided for the type of workload/container this best practice applies to. The context, of an individual best practice, should illustrate the specific type of workload most likey to take advantage of the benefits listed. In contrast, it should provide a warning if the practice would be a determent to a divergent type of workload.
```
taylor:
```suggestion
Context should be provided which includes the type of CNF, workload, and CNF component (eg. Pod, container, Operator) this best practice applies to. The context, of an individual best practice, should illustrate the specific type of workload most likely to take advantage of the benefits listed. When applicable, it should also provide a warning if the practice would be a determent to a divergent type of workload.
```

taylor (changes requested):
@jeffsaelens seems like the terminology could support "infra" and "application" CNFs.

ccdps/0001-cnf-conformance-definition-proposal-process.md
jeffsaelens: Sure, infrastructure/orchestration/provisioning style workloads. One that we do actually want to have some type of privilege as they are interacting with systems on our behalf. The core network functions themselves. These ones may need no privilege but have unique dos/don'ts from a NAT, ingress/egress etc... perspective. OSS/BSS containers, logging/monitoring (most likely to be bundled in by the dev to ensure the appropriate metrics are actually captured). The list could go on. One issue is we don't have a proper definition of what a CNF is. Is it a single container that has nay sort of NF living inside? Is it a suite of microservices that make up the whole? My main fear is we describe some of these best practices as one size fits all and end up with ETSI MANO again where things are a little too broad and begin to lose value.

taylor: @jeffsaelens these examples would be a good addition to the document.  Please add a few then we can merge this PR.

We'll address what a CNF is in some of the primary docs (eg. README, charter) partially based on discussions in https://github.com/cncf/cnf-wg/discussions/27 and recent ones in Slack.

taylor:
@rabi-abdel, @vukg

tomkivlin (approved):
lgtm

xmulligan:
@fkautz what do you think?

taylor (approved):

xmulligan:
From 2021-01-25 meeting
[Shane] is it a single container or multiple
[Jambi] Should this be a drop down vs a free form?

PR #46, opened 2021-01-25
use case folder and README

taylor:
## How to contribute use cases

This adds a use case folder and instructions on contributing
- add to list https://github.com/cncf/cnf-wg/discussions/39
- create new discussion topic
- create PRs

### DIFF
--- a/use-case/README.md
+++ b/use-case/README.md
@@ -6,7 +6,7 @@ This folder contains different use cases, their description, references, and oth
 
 Goal:
 
-- Relating best practices to real world use cases.
+- Relating to real world use cases.
 - Discussion of core features and functionality from use cases and their NFs.
 - Examination of existing use cases to discussion of kubenative implementations which provide the core features and functionality.
 
--- a/use-case/README.md
+++ b/use-case/README.md
@@ -1,6 +1,6 @@
 # Networking Use Cases
 
-Proposed best practices are rooted in our model of how CNFs are used.  A best practice is most clearly the right thing to do if it helps with a use case related to one or more of the actors.
+It's clearer that a best practice is a good course of action when it satisfies the needs of a Kubernetes use case related to one or more actors.```
 
 This folder contains different use cases, their description, references, and other context.  The best practices, recorded elsewhere, refer back to these use cases for their justification.
 
--- a/use-case/README.md
+++ b/use-case/README.md
@@ -14,7 +14,7 @@ Also see related discussion topic https://github.com/cncf/cnf-wg/discussions/39
 
 ## How to add a new use case
 
-If you just have an idea for a use case feel free to add to the discussion topic [listing networking uses cases](https://github.com/cncf/cnf-wg/discussions/39)
+If you just have a part-formed idea for a use case feel free to add to the discussion topic [listing networking uses cases](https://github.com/cncf/cnf-wg/discussions/39).  This is the best place to go if you aren't sure how to write up the use case.  We can help you expand that into a full description.
 
 If you have a new use case you would like to discuss in more depth, create a new [github discussion topic](https://github.com/cncf/cnf-wg/discussions) with the details you have including attachments and/or links to any external content (eg. Google docs, hackmd, diagrams).
 
--- a/use-case/README.md
+++ b/use-case/README.md
@@ -14,7 +14,7 @@ Also see related discussion topic https://github.com/cncf/cnf-wg/discussions/39
 
 ## How to add a new use case
 
-If you just have a part-formed idea for a use case feel free to add to the discussion topic [listing networking uses cases](https://github.com/cncf/cnf-wg/discussions/39).  This is the best place to go if you aren't sure how to write up the use case.  We can help you expand that into a full description.
+If you just have a partially-formed idea for a use case feel free to add to the discussion topic [listing networking uses cases](https://github.com/cncf/cnf-wg/discussions/39).  This is the best place to go if you aren't sure how to write up the use case.  We can help you expand that into a full description.
 
 If you have a new use case you would like to discuss in more depth, create a new [github discussion topic](https://github.com/cncf/cnf-wg/discussions) with the details you have including attachments and/or links to any external content (eg. Google docs, hackmd, diagrams).
 
--- a/use-case/README.md
+++ b/use-case/README.md
@@ -1,0 +1,23 @@
+# Networking Use Cases
+
+As part of any proposed best practices it is requested to provide 1 or more use cases showing where the best practice could be used.
+
+This folder contains different networking use cases, their description, example references, and other context for each.
+
+Goal:
+
+- Relating best practices to real world use cases.
+- Discussion of core features and functionality from use cases and their NFs.
+- Examination of existing use cases to discussion of kubenative implementations which provide the core features and functionality.
+
+Also see related discussion topic https://github.com/cncf/cnf-wg/discussions/39
+
+## How to add a new use case
+
+If you just have an idea for a use case feel free to add to the discussion topic [listing networking uses cases](https://github.com/cncf/cnf-wg/discussions/39)
+
+If you have a new use case you would like to discuss, create a new [github discussion topic](https://github.com/cncf/cnf-wg/discussions) with the details you have including attachements and/or links to any external content (eg. Google docs, hackmd, diagrams).
+
+If you have enough content to writeup a full use case then feel free to create a pull request in this [use case folder](https://github.com/cncf/cnf-wg/tree/master/use-case) with either a
+- single markdown file with all content included (including in-line diagrams) for the use case
+- new use case folder with all content with a README.md as the starting point describing the use case
--- a/use-case/README.md
+++ b/use-case/README.md
@@ -1,6 +1,6 @@
 # Networking Use Cases
 
-As part of any proposed best practices it is requested to provide 1 or more use cases showing where the best practice could be used.
+Proposed best practices are rooted in our model of how CNFs are used.  A best practice is most clearly the right thing to do if it helps with a use case related to one or more of the actors.
 
 This folder contains different use cases, their description, references, and other context.
 
--- a/use-case/README.md
+++ b/use-case/README.md
@@ -20,4 +20,6 @@ If you have a new use case you would like to discuss in more depth, create a new
 
 If you have enough content to writeup a full use case then feel free to create a pull request in this [use case folder](https://github.com/cncf/cnf-wg/tree/master/use-case) with either a
 - single markdown file with all content included (including in-line diagrams) for the use case
-- new use case folder with all content with a README.md as the starting point describing the use case
+- new use case folder with all content with a README.md as the starting point describing the use case.
+
+if you're going to do this via a pull request you must have a complete description of the use case.   You should have gathered any information you need and included it by this point.  We will then discuss the merits of the use case during review of the pull request, but this is not the time to be writing large sections of it.
--- a/use-case/README.md
+++ b/use-case/README.md
@@ -2,7 +2,7 @@
 
 As part of any proposed best practices it is requested to provide 1 or more use cases showing where the best practice could be used.
 
-This folder contains different networking use cases, their description, example references, and other context for each.
+This folder contains different use cases, their description, references, and other context.
 
 Goal:
 
--- a/use-case/README.md
+++ b/use-case/README.md
@@ -16,7 +16,7 @@ Also see related discussion topic https://github.com/cncf/cnf-wg/discussions/39
 
 If you just have an idea for a use case feel free to add to the discussion topic [listing networking uses cases](https://github.com/cncf/cnf-wg/discussions/39)
 
-If you have a new use case you would like to discuss, create a new [github discussion topic](https://github.com/cncf/cnf-wg/discussions) with the details you have including attachements and/or links to any external content (eg. Google docs, hackmd, diagrams).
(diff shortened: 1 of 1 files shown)


xmulligan:
Do we want to be anymore prescriptive on how people should submit their use cases?

taylor:
> Do we want to be anymore prescriptive on how people should submit their use cases?

I didn't want to slow down contributions -- people have communicated an interest recently -- so I did not add a template or more prescriptive instructions.  If there is a useful template someone can contribute that. Otherwise we can organically become more prescriptive as use cases are submitted if needed/desired.

use-case/README.md
```diff
@@ -0,0 +1,23 @@
+# Networking Use Cases
+
+As part of any proposed best practices it is requested to provide 1 or more use cases showing where the best practice could be used.
```
iawells: No need to hedge in an instruction document.  If this document is accepted, it's no longer a suggestion.
```suggestion
Proposed best practices are rooted in our model of how CNFs are used.  A best practice is most clearly the right thing to do if it helps with a use case related to one or more of the actors.
```

use-case/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+This folder contains different networking use cases, their description, example references, and other context for each.
```
iawells: I know we're thinking about networking, but things like usability also deserve their time in the sun.
```suggestion
This folder contains different use cases, their description, references, and other context.
```

use-case/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Relating best practices to real world use cases.
+- Discussion of core features and functionality from use cases and their NFs.
+- Examination of existing use cases to discussion of kubenative implementations which provide the core features and functionality.
+
+Also see related discussion topic https://github.com/cncf/cnf-wg/discussions/39
+
+## How to add a new use case
+
+If you just have an idea for a use case feel free to add to the discussion topic [listing networking uses cases](https://github.com/cncf/cnf-wg/discussions/39)
```
iawells:
```suggestion
If you just have a part-formed idea for a use case feel free to add to the discussion topic [listing networking uses cases](https://github.com/cncf/cnf-wg/discussions/39).  This is the best place to go if you aren't sure how to write up the use case.  We can help you expand that into a full description.
```

use-case/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+If you have a new use case you would like to discuss, create a new [github discussion topic](https://github.com/cncf/cnf-wg/discussions) with the details you have including attachements and/or links to any external content (eg. Google docs, hackmd, diagrams).
```
iawells:
```suggestion
If you have a new use case you would like to discuss in more depth, create a new [github discussion topic](https://github.com/cncf/cnf-wg/discussions) with the details you have including attachments and/or links to any external content (eg. Google docs, hackmd, diagrams).
```

use-case/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+If you have enough content to writeup a full use case then feel free to create a pull request in this [use case folder](https://github.com/cncf/cnf-wg/tree/master/use-case) with either a
+- single markdown file with all content included (including in-line diagrams) for the use case
+- new use case folder with all content with a README.md as the starting point describing the use case
```
iawells: OK, we're going to want a template for this with basic points we need to see in a use case.

The problem is that we have both use cases and user stories here and I see I'm already confusing the two: a use case is probably more about 'this is what we already do in the world' and a user story is more 'a named actor needs to get from point A to point B'.  I don't see much point in separating them - they both underpin our recommendations - but it might mean we need two templates to choose from.

Can I draw a line between 'external content' and 'citations'?  If we're talking about a preliminary writeup in google docs, then that's 'external content' - it might go away, but it's optional.  However, diagrams crucial to the explanation want committing into this repo as they need change tracking.  Citations are also external references demonstrating that what's written is objectively correct (rather than someone's opinion), and they do (as far as possible) want to be permalinks, fixed up if they break.

iawells (changes requested):

use-case/README.md
taylor: @iawells partially-formed vs part-formed?  I like the suggestion otherwise.

use-case/README.md
taylor: @iawells I like the idea of 2 templates. Can we add those after this PR is merged?  If we have a PR before templates are created we can always either delay the merge or merge and update the use case to fit the new template(s) when they are created.

Also see https://github.com/cncf/cnf-wg/pull/46#issuecomment-767721540

taylor: I agree that citations can be external, but context crucial to the use case or user story (committed to the repo) should have all content included (I count GitHub's CDN as included)

use-case/README.md
```diff
@@ -0,0 +1,23 @@
+# Networking Use Cases
+
+Proposed best practices are rooted in our model of how CNFs are used.  A best practice is most clearly the right thing to do if it helps with a use case related to one or more of the actors.
+
+This folder contains different use cases, their description, references, and other context.
+
+Goal:
+
+- Relating best practices to real world use cases.
```
electrocucaracha: As it was mentioned before best practices provides well-known methods to solve a specific problem. Having a clear distinction between the problem and the solution can facilitate its understanding. So maybe separating this bullet can facilitate that.
```suggestion
- Relating real world use cases.
- Best practices proposed.
```

taylor: @electrocucaracha understood. This README is talking about goals for the use cases.  So best practices stand-alone does not make sense to me here.

It could be reworded though... hmmm

taylor: @electrocucaracha how about if we just delete the entire goals section? The wording before it seems to cover the intent.

taylor, line 9:
```suggestion
```

Delete ^^^

iawells: The aim is to document best practices here so that we can relate use cases to them (in the future, from the other part of the repo).  Something spun along that line?

use-case/README.md
```diff
@@ -0,0 +1,23 @@
+# Networking Use Cases
+
+Proposed best practices are rooted in our model of how CNFs are used.  A best practice is most clearly the right thing to do if it helps with a use case related to one or more of the actors.
+
+This folder contains different use cases, their description, references, and other context.
```
iawells:
```suggestion
This folder contains different use cases, their description, references, and other context.  The best practices, recorded elsewhere, refer back to these use cases for their justification.
```

use-case/README.md
taylor: Originally I added the "Goal" section to communicate what the purpose of these use cases were. It seems like it is redundant and not needed right now.

I suggest deleting it at this point https://github.com/cncf/cnf-wg/pull/46/files#r565533928

use-case/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+## How to add a new use case
+
+If you just have a partially-formed idea for a use case feel free to add to the discussion topic [listing networking uses cases](https://github.com/cncf/cnf-wg/discussions/39).  This is the best place to go if you aren't sure how to write up the use case.  We can help you expand that into a full description.
+
+If you have a new use case you would like to discuss in more depth, create a new [github discussion topic](https://github.com/cncf/cnf-wg/discussions) with the details you have including attachments and/or links to any external content (eg. Google docs, hackmd, diagrams).
+
+If you have enough content to writeup a full use case then feel free to create a pull request in this [use case folder](https://github.com/cncf/cnf-wg/tree/master/use-case) with either a
+- single markdown file with all content included (including in-line diagrams) for the use case
+- new use case folder with all content with a README.md as the starting point describing the use case
```
iawells:
```suggestion
- new use case folder with all content with a README.md as the starting point describing the use case.

if you're going to do this via a pull request you must have a complete description of the use case.   You should have gathered any information you need and included it by this point.  We will then discuss the merits of the use case during review of the pull request, but this is not the time to be writing large sections of it.
```

iawells (changes requested):

use-case/README.md
taylor: @iawells @electrocucaracha the purpose of the "Goal" section was to communicate why we were creating/documenting these use cases 

> Goal:
> 
> - Relating best practices to real world use cases.
> - Discussion of core features and functionality from use cases and their NFs.
> - Examination of existing use cases to discussion of kubenative implementations which provide the core features and functionality.

The first bullet point, "Relating best practices to real world use cases," is related to the description update you suggested @iawells 
> A best practice is most clearly the right thing to do if it helps with a use case related to one or more of the actors.


Maybe the first bullet point can be reworded to something like:

> Provide use cases for best practices to reference

iawells:
I am now a happy bunny.

use-case/README.md
iawells: Right - we have our plans for templates, but we won't put them in yet.  Let's accept some use case pull requests and use that to feed what the template needs to look like.

use-case/README.md
iawells: If you took the other change I think it's covered and you can drop it entirely, but your suggestions are fine.

electrocucaracha: @taylor agree, the main idea is to provide uses cases and document their best practices used, isn't it? but maybe they can be categorized by the problem that they're solving and help to support those best practices. So  adding information about the problem provides more context about the decision of choosing a set of best practices, e.g. "This use case requires high availability without low latency"

taylor: @electrocucaracha I'm up for having just

- Relating real world use cases.


and not using 

- Best practices proposed.

as a change in goals for the use case readme. The latter does not make sense stand alone to me.  (unfinished sentence/thought)

taylor:
> I am now a happy bunny.

@iawells great :)

Now mark the PR as approved ✔️ 🐰

use-case/README.md
```diff
@@ -0,0 +1,25 @@
+# Networking Use Cases
+
+Proposed best practices are rooted in our model of how CNFs are used.  A best practice is most clearly the right thing to do if it helps with a use case related to one or more of the actors.
+
+This folder contains different use cases, their description, references, and other context.  The best practices, recorded elsewhere, refer back to these use cases for their justification.
```
rabi-abdel, line 5: @taylor i think it will be useful to clearly specify what information we need to see in a use case, in a bullet list.

- Problem statment.
- impact to real-life actor.
-  etc.

rabi-abdel (commented):
looks good.

use-case/README.md
```diff
@@ -0,0 +1,25 @@
+# Networking Use Cases
+
+Proposed best practices are rooted in our model of how CNFs are used.  A best practice is most clearly the right thing to do if it helps with a use case related to one or more of the actors.
```
jeffsaelens: The second sentence reads awkwardly. "Most Cleary" is very strongly worded. It sounds almost biblical... Can we go a little softer and pull in our subject domain?

"A best practice is agreed upon as the optimal course of action when it satisfies the needs of a Kubernetes use case related to one or more actors?" This isn't perfect, but it has a little less finality in it.

use-case/README.md
jeffsaelens: To Rabi's point, at some point we'd probably want a template for use cases and a means of pointing back to them from the best practices (this could also be a one to many scenario so some thought might be needed). I know we are trying to keep people's creative spirits soaring, but we might want more specificity as to when to use "discussions" for more free form thoughts on paper versus when a drafted proposal is appropriate.

jeffsaelens (changes requested):
Makes sense overall, I think a little tweaking is in order. Additionally, after this PR (no need to bog this one down), we should either put a template in the directory, or add additional parameters to the README as suggested by others. There is parts in the README that call out the discussions section, but its somewhat fuzzy of when you post in in there vs drafting a doc here.

use-case/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+This folder contains different use cases, their description, references, and other context.  The best practices, recorded elsewhere, refer back to these use cases for their justification.
+
+Goal:
+
+- Relating best practices to real world use cases.
```
xmulligan:
```suggestion
- Relating to real world use cases.
```

use-case/README.md
iawells:
```suggestion
It's clearer that a best practice is a good course of action when it satisfies the needs of a Kubernetes use case related to one or more actors.```

use-case/README.md
xmulligan: We will add the template late. Issue to be created

taylor: @rabi-abdel @jeffsaelens can we create an issue to address this?

iawells (changes requested):

use-case/README.md
taylor: @jeffsaelens verbally said yes on the CNF WG call

@xmulligan will create a new issue

iawells:
I would focus on:

Organisations involved
- SP, platform vendor, telco vendor (the SP may not use vendors or they may), SI, ...

Roles:
- Network architect - planning out the network
- CNF dev - writes a component they would want to use
- Platform dev - writes an operable (i.e. deployable and upgradeable) platform
- SI - glues the CNF and platform network together per architect's design and maybe patches up things the operator requests afterward
- Operator - day to day person that runs the resulting network

Then a section per role; and include in that section a discussion of the perspectives the role brings (e.g. reliability, security).

That's less fine grained than Taylor has in actors but the perspectives should cover the rest.

Then some example assignments of roles to orgs and additional considerations for that sort of breakdown (e.g. communications between roles, purchasing, when CNF is bought in, implications of an RFP process) - this comes last because it's the one that diverts us from the more technical details.

Issue #54, opened 2021-02-02
Define Actors and their roles 


xmulligan:
There is a lot of content under discussion #30 for defining the roles, this should now be written up into the first draft of a PR

xmulligan:
@iawells
we are missing Roles
Network architect - planning out the network
CNF dev - writes a component they would want to use
Platform dev - writes an operable (i.e. deployable and upgradeable) platform
SI - glues the CNF and platform network together, with configuration and either written processes or orchestration, per architect's design; and maybe patches up things the operator requests afterward
Operator - day to day person that runs the resulting network


Ian: Need to start working towards a document to add to the repo
Taylor: [Shared google doc](https://docs.google.com/document/d/1cHaGXr52e3g_3CHXGdwcBhzWEhRfpIu-MDp7MQWEZeg/edit#heading=h.65ryuzju21y3)

taylor:
@iawells provided some information on roles in https://github.com/cncf/cnf-wg/discussions/30#discussioncomment-328034

xmulligan:
@iawells raised his hand and putting it under a top level folder

xmulligan:
Closing this due to inactivity. It can be reopened at a later time if relevant again.

PR #72, opened 2021-02-23
added use case template & referenced it in readme

### LINKED ISSUE #53: Create template for use cases
From #46 https://github.com/cncf/cnf-wg/pull/46#discussion_r566053891

@rabi-abdel 
 i think it will be useful to clearly specify what information we need to see in a use case, in a bullet list.

Problem statment.
impact to real-life actor.
etc.

@jeffsaelens 
To Rabi's point, at some point we'd probably want a template for use cases and a means of pointing back to them from the best practices (this could also be a one to many scenario so some thought might be needed). I know we are trying to keep people's creative spirits soaring, but we might want more specificity as to when to use "discussions" for more free form thoughts on paper versus when a drafted proposal is appropriate.

vukg:
Hi There,

Here is use case template with references in CCDP readme that shall resolve #53. It is first draft for the discussion. If you want to merge it in the feature branch please create one and let me know so I will change the pull request.

Looking forward to get some comments and sorry for waiting on it.

-Vuk


ccdps/NNNN-UC-template.md
```diff
@@ -0,0 +1,41 @@
...
+- [ ] Development
+- [ ] Deployment
+- [ ] System integration
+- [ ] Network integration
+- [ ] Lifecycle management
+- [ ] Operations
+
+## Involved parties
+*TIP: In this section you can list main parties who are involved in the use case. For example:<br>
+Business customer of CSP<br>
```
iawells: So the we use 'roles' in agile and that helps a little:

1. Sometimes one person has multiple roles, so it saves the 'jack of all trades does X' kind of user story
2. We don't have to assign roles to companies, for instance, which helps when @jeffsaelens develops his own CNFs despite being an operator

I like your breakdown in general, but I wonder if we can rethink the first list to be roughly around the role list that @taylor was creating - maybe a bit reduced - and fold in the second one.  If we want to talk about 'X buys a CNF from an external party' that is something we could do as a user story by itself, theoretically.

ccdps/NNNN-UC-template.md
```diff
@@ -0,0 +1,41 @@
...
+Operations Team of CSP<br>
+Service team of CNF Vendor*
+
+## Involved system entities
+*TIP: In this section you can list system entities that are involved in the use case. For example:<br>
+Data Center Network<br>
+Wide Area Network<br>
+Cloud Native Platform<br>
+CNF X<br>
```
iawells: I really like this bit.  I would add 'MANO' and 'OSS'.

ccdps/NNNN-UC-template.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Situation
```
iawells: These two sections (situation/expected behaviour) are probably just one option for telling the story.  It raises the question of 'must you use this template or is it just a helpful place to start?'

ccdps/NNNN-UC-template.md
```diff
@@ -0,0 +1,41 @@
...
+*TIP: Please describe situation related to this use case. Indicate what are intents and expectations of different parties as well as the roles and requirements of different system components. Describe how everything is related. Drawings, flow and system diagrams are beneficial here. This part, as the subject says, shall give the reader clear overview of the situation as the starting point for evaluation of the use case. It should not contain the detailed technical descriptions and analysis.*
+
+## Expected behaviour
+*TIP: Here you can provide detailed technical descriptions and depict expected behaviour of system entities and components regardless of cloud nativeness. Please highlight what are different entities expected to provide. Depending on the use case you can look at it from different angles e.g. CNF angle, Cloud Native Platform angle, Network angle etc. This section should give the reader clear understanding of expected behaviour.
+
+## Challenges and limitations with Kube-native approach
```
iawells: Seems reasonable.  Misses out the 'this is the platform I am measuring against' that we had in the first user story, but we were talking about whether to extract that (in which case we can reference it here).

ccdps/NNNN-UC-template.md
```diff
@@ -0,0 +1,41 @@
+# UC-NNNN: Short descriptive title of your use case (TEMPLATE)
+
```
iawells: The first user story got a glossary, by request.  Should we suggest one?

ccdps/NNNN-UC-template.md
claudiobartolini: Agree with the above comment. May want to use actors as a first-class concept here and cross reference the part where those can be defined.

claudiobartolini (commented):
Suggest to have "intended benefits" or something along those lines made explicit. (Anything that answers the "why?" question justifying the use case

ccdps/NNNN-UC-template.md
xmulligan: I agree with @claudiobartolini, I think we should define the actors in separate part and solve #54 . This sections can then cross reference that document and we can keep/update them separatley

ccdps/README.md
```diff
@@ -8,3 +8,6 @@ This process is still in a alpha state and feedback is welcome.
 Socialize an idea with the CNF WG. You can send your idea to the mailing list, or add it to the agenda for one of the upcoming meetings. Make sure that others think the work is worth taking up and will help review the CCDP and any code changes required.
 
 Follow the process outlined in the [CCDP process document](0001-cnf-conformance-definition-proposal-process.md) and use the [CCDP template](NNNN-ccdp-template.md) to create a new proposal. Submit a pull request to add the proposal to the repo.
+
+## Use Cases
+CCDP shall always refer to at least one of CNF WG use cases. Those use cases serve as the practical and concrete basis for deriving, discussing and evaluating the best practices for CNFs. Their value is in providing context and sanity check for CCDPs. With the help of use cases we are making sure that CCDPs are addressing specific problem or need from the real world. Please use [UC-template.md](Use Case Template) to contribute an use case.
```
xmulligan:
```suggestion
CCDP shall always refer to at least one of CNF WG use cases. Those use cases serve as the practical and concrete basis for deriving, discussing, and evaluating the best practices for CNFs. Their value is in providing context and sanity check for CCDPs. With the help of use cases, we are making sure that CCDPs are addressing specific problems or needs from the real world. Please use [Use Case Template](UC-template.md) to contribute an use case.
```

xmulligan:
I also think this should be in the folder created in #46 rather than in the ccdp folder

ccdps/NNNN-UC-template.md
vukg: I was actually having human actors in mind or even personas. I wanted to leave it a bit freestyle to give degree of freedom as we might miss some if we define it as a "foreign key". If you like more to keep actors and cross reference I will change it like that. Otherwise I could go with personas. WDYT?

ccdps/NNNN-UC-template.md
vukg: I have added more examples of entities.

ccdps/NNNN-UC-template.md
vukg: Situation is rather to describe point of departure - explaining the scene. Expected behavior is describing how do we expect "the plot" to unfold. If we want to keep comparable structure of use cases I would opt for "must use" template.

ccdps/NNNN-UC-template.md
iawells: Whether persona or role, we seem to be settling on the same point; that we should be talking about the hat they're wearing at the time, as opposed to e.g. 'architect employed by a telco' who might wear many hats depending on the telco's internal structure, where they're getting their CNFs from, and so on.

ccdps/NNNN-UC-template.md
iawells: I know what you're trying to say, but if you're telling a user story it's usually a stepwise process - 'wants X to happen' then 'does X, expects Y' over and over.  Use cases work out differently because they're more about objects than people (see 'BGP server').  I'm not sure I see how we could make this apply quite right to either of those.

ccdps/NNNN-UC-template.md
vukg: Agreed. I will add it.

iawells:
> Suggest to have "intended benefits" or something along those lines made explicit. (Anything that answers the "why?" question justifying the use case

OK, but use cases are trying to be the 'why' for the best practices (and I think the best practices are the things that have specific benefits).  Using the BGP use case as an example, how would you frame a 'why' for it?

I wonder if this is more a question of expressing the value of being able to do what the use case describes?

ccdps/NNNN-UC-template.md
vukg: Situation actually tries to capture intents/needs of actors - understanding business/operational context if you will. Expected behavior is more about how, given the intents/needs, of actors the objects should behave.

ccdps/NNNN-UC-template.md
jeffsaelens: I like Ian's hat approach. As we see with focused/targeted calls, a lot of the usual suspects show up as they fill multiple roles. The industry tends to have this trend as well. Its not uncommon to have someone who's been on with a vendor, a telco and on ops and eng throughout their career.

ccdps/README.md
```diff
@@ -8,3 +8,6 @@ This process is still in a alpha state and feedback is welcome.
 Socialize an idea with the CNF WG. You can send your idea to the mailing list, or add it to the agenda for one of the upcoming meetings. Make sure that others think the work is worth taking up and will help review the CCDP and any code changes required.
 
 Follow the process outlined in the [CCDP process document](0001-cnf-conformance-definition-proposal-process.md) and use the [CCDP template](NNNN-ccdp-template.md) to create a new proposal. Submit a pull request to add the proposal to the repo.
+
+## Use Cases
+CCDP shall always refer to at least one of CNF WG use cases. Those use cases serve as the practical and concrete basis for deriving, discussing, and evaluating the best practices for CNFs. Their value is in providing context and sanity check for CCDPs. With the help of use cases, we are making sure that CCDPs are addressing specific problems or needs from the real world. Please use [Use Case Template](NNNN-UC-template.md) to contribute an use case.
```
xmulligan:
```suggestion
CCDP shall always refer to at least one of CNF WG use case or user story. Those use cases serve as the practical and concrete basis for deriving, discussing, and evaluating the best practices for CNFs. Their value is in providing context and sanity check for CCDPs. With the help of use cases, we are making sure that CCDPs are addressing specific problems or needs from the real world. Please use [Use Case Template](NNNN-UC-template.md) to contribute an use case.
```

taylor: The section header needs to say user story as well

`## Use Cases and User Stories`

ccdps/NNNN-UC-template.md
iawells: OK, so per the meeting we can have a hat list here with a reasonable list of suggestions and we'll break that out into an independent list with our considered opinions when we've considered @taylor's work more in-depth - that will be a separate commit.

claudiobartolini (approved):

ccdps/NNNN-UC-template.md
iawells: I suggest we commit with this and how if it works in practice.  Then we will have hard evidence for whether it needs changing rather than personal opinions.

ccdps/README.md
iawells: Or just 'requirements docs'; perhaps with a reference to the folder.

ccdps/NNNN-UC-template.md
```diff
@@ -0,0 +1,48 @@
+# UC-NNNN: Short descriptive title of your use case (TEMPLAT)
+
+## Glossary
+*TIP: Please list the key technical terms and acronyms and provide definitions for those.*
```
electrocucaracha:
```suggestion
> Provide definitions for technical terms and acronyms.
```

ccdps/NNNN-UC-template.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Involved processes
+*TIP: Please mark key processes that are involved in the use case you are presenting. This helps actors to decide if the use case is relevant for them or not. If there are more processes involved please make multiple selection.*
```
electrocucaracha: Another option can be to uncomment the processes like
```suggestion
> Please uncomment and/or add key processes that are involved in the use case you are presenting. This helps actors to decide if the use case is relevant for them or not.
> - Development
> - Deployment
> - System integration
> - Network integration
> - Lifecycle management
> - Operations
```

ccdps/NNNN-UC-template.md
```diff
@@ -0,0 +1,48 @@
...
+- [ ] Development
+- [ ] Deployment
+- [ ] System integration
+- [ ] Network integration
+- [ ] Lifecycle management
+- [ ] Operations
+
+## Involved parties
+*TIP: In this section you can list main parties who are involved in the use case. For example:<br>
```
electrocucaracha:
```suggestion
> List main parties involved in the use case. For example:
```

ccdps/NNNN-UC-template.md
```diff
@@ -0,0 +1,48 @@
...
+Business customer of CSP<br>
+Operations Team of CSP<br>
+Service team of CNF Vendor<br>
+Etc.*
+
+## Involved system entities
+*TIP: In this section you can list system entities that are involved in the use case. For example:<br>
```
electrocucaracha:
```suggestion
> List system entities that are involved in the use case. For example:
```

ccdps/NNNN-UC-template.md
```diff
@@ -0,0 +1,48 @@
...
+Data Center Network<br>
+Wide Area Network<br>
+Cloud Native Platform<br>
+CNF X<br>
+MANO<br>
+OSS<br>
+Etc.*
+
+## Situation
+*TIP: Please describe situation related to this use case. Indicate what are intents and expectations of different parties as well as the roles and requirements of different system components. Describe how everything is related. Drawings, flow and system diagrams are beneficial here. This part, as the subject says, shall give the reader clear overview of the situation as the starting point for evaluation of the use case. It should not contain the detailed technical descriptions and analysis.*
```
electrocucaracha:
```suggestion
> Describe situation related to this use case. Indicate what are intents and expectations of different parties as well as the roles and requirements of different system components. Explain how everything is related. Drawings, flow and system diagrams are beneficial here. This part shall give the reader clear overview of the situation as the starting point for evaluation of the use case. It should not contain the detailed technical descriptions and analysis.
```

ccdps/NNNN-UC-template.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Expected behaviour
+*TIP: Here you can provide detailed technical descriptions and depict expected behaviour of system entities and components regardless of cloud nativeness. Please highlight what are different entities expected to provide. Depending on the use case you can look at it from different angles e.g. CNF angle, Cloud Native Platform angle, Network angle etc. This section should give the reader clear understanding of expected behaviour.
```
electrocucaracha:
```suggestion
behaviour or behavior?
> Provide detailed technical descriptions and depict expected behaviour of system entities and components regardless of cloud nativeness. Highlight what are different entities expected to provide. Depending on the use case you can look at it from different angles e.g. CNF angle, Cloud Native Platform angle, Network angle etc. This section should give the reader clear understanding of expected behaviour.
```

ccdps/NNNN-UC-template.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Challenges and limitations with Kube-native approach
+*TIP: This section will give us main inputs for discussion and evaluation of best practices. You should describe which challenges and limitations kube-native approach presents for this use case. You can give technical details about what is not possible with kube-native approach and elaborate why is that relevant. You can again look from multiple angles. If there are some established practices that are used to work around these limitations please describe those in the chapter below*
```
electrocucaracha:
```suggestion
> This section will give us main inputs for discussion and evaluation of best practices. Describe which challenges and limitations kube-native approach presents for this use case. Provide technical details about what is not possible with kube-native approach and elaborate why is that relevant. If there are some established practices that are used to work around these limitations please describe those in the chapter below*
```

ccdps/NNNN-UC-template.md
```diff
@@ -0,0 +1,48 @@
...
+
+### Established practices to overcome challenges and limitations
+*TIP: If such practices exist please elaborate them here in sufficient technical details. Please provide refences. Please give your opinion weather or not these practices can be considered as candidates for CNF best practices in line with [CCDP description](0001-cnf-conformance-definition-proposal-process.md).*
```
electrocucaracha:
```suggestion
> If such practices exist please elaborate them here in sufficient technical details. Provide references and give your opinion weather or not these practices can be considered as candidates for CNF best practices in line with [CCDP description](0001-cnf-conformance-definition-proposal-process.md).
```

ccdps/NNNN-UC-template.md
```diff
@@ -0,0 +1,48 @@
...
+
+### What needs to be done differently in order to overcome challenges and limitations 
+*TIP: If there are no established practices and you have ideas or if you believe things related to this use case should be done differently please indicate it shortly here. The purpose of this is to earmark some potential inputs for CCDP process. It is not expected to elaborate CCDPs here.
```
electrocucaracha:
```suggestion
> If there are no established practices and you have ideas or if you believe things related to this use case should be done differently please indicate it shortly here. The purpose of this is to earmark some potential inputs for CCDP process. It is not expected to elaborate CCDPs here.
```

electrocucaracha (commented):
My suggestions are more related to the format. @vukg if you can remove breakline html tags should be nice.

ccdps/NNNN-UC-template.md
sronanrh: The way this is deal with elsewhere, is a single person can fill two roles, but that does not combined the roles, they still exists as distinct roles.

ccdps/NNNN-UC-template.md
```diff
@@ -0,0 +1,48 @@
...
+OSS<br>
+Etc.*
+
+## Situation
+*TIP: Please describe situation related to this use case. Indicate what are intents and expectations of different parties as well as the roles and requirements of different system components. Describe how everything is related. Drawings, flow and system diagrams are beneficial here. This part, as the subject says, shall give the reader clear overview of the situation as the starting point for evaluation of the use case. It should not contain the detailed technical descriptions and analysis.*
+
+## Expected behaviour
+*TIP: Here you can provide detailed technical descriptions and depict expected behaviour of system entities and components regardless of cloud nativeness. Please highlight what are different entities expected to provide. Depending on the use case you can look at it from different angles e.g. CNF angle, Cloud Native Platform angle, Network angle etc. This section should give the reader clear understanding of expected behaviour.
+
+## Challenges and limitations with Kube-native approach
```
rannyh: Is 'Kube-Native' defined somewhere? I thought we were using the term 'Cloud-Native'. Is it the same?

iawells: I think by using kube-native, we're saying 'designed to run specifically on Kubernetes'.

`<rant>`

The face-value definition of 'cloud native' is 'designed to run on a cloud rather than in some other way'. 
 Arguably that might be a virtual machine based cloud, and it certainly doesn't have to be a container based cloud.

You'll note that the CNCF cloud native definition - https://github.com/cncf/toc/blob/main/DEFINITION.md - doesn't actually define cloud native; instead, it sort of says 'you'll know it when you see people using techniques that make it easier to build cloud-native software'.

Since it's a surprisingly woolly term, and people don't consistently mean the same thing when they use it, we should be quite cautious about clarifying what we mean when it comes up.

A fine example of the term being abused is when we talk about a 'cloud native platform' - it doesn't make much sense to refer to cloud platforms as 'cloud native'.

`</rant>`

electrocucaracha: @iawells I'm totally agree with that distinction but I guess what @rannyh is referring is the fact that Kube-Native has not been documented anywhere in this repo and eventually can be used by other documents.

rannyh: Yeah. I was merely pointing out that the term "kube-native" was not defined anywhere in this document set. Didn't mean to start a debate about the meaning of cloud-nativeness. So going back to my question - are we going to use "kube-native" and add a definition, or use "cloud-native" like in the rest  of this document set?

vukg: I have specifically used kube-native there and expected to trigger some discussion with it. I also recall that @taylor was using similar term. Having said that, I believe we need to be focused and true to ourselves and focus on Network Functions that run in Kubernetes cluster + necessary components in and around that cluster.

Cloud native platform or infrastructure is actually real thing. It is infrastructure that exhibits the properties of a cloud and that enables cloud native applications to run on it. Take example of Kubernetes cluster deployed on 10 bare metal nodes on the edge: it can not be called cloud, but it is cloud native infrastructure as it exhibits all properties of cloud (no HA of individual nodes, API driven etc). One more generalized reference for cloud native infrastructure: https://books.google.de/books/about/Cloud_Native_Infrastructure.html?id=1Fk7DwAAQBAJ&printsec=frontcover&source=kp_read_button&redir_esc=y#v=onepage&q&f=false

electrocucaracha: @vukg I'm agree with that so maybe the missing part is the addition of the term into glossary's documentation.

vukg:
Hey all,
I have made number of corrections to the template in line with the discussions. I have also added one example use case according to the template: https://github.com/cncf/cnf-wg/blob/29202cfd67f6ca2f0175a0e53b51d70ae6529cc8/use-case/0001-UC-lifecycle-of-infrastructure-where-CNF-is-running.md.

Please let review and comment.

Vuk

xmulligan (approved):

iawells:
This pull request doesn't include the BGP use case, but if you look there we had a subdirectory with a README file in it with body text and the directory could be used for attachments.  I think that's wise; can we stick with that format?

I didn't number the use case I added.  I'm not sure for the reason why we would  number use cases and CCDPs (it's in a directory, the name will be unique) but I presume whoever made the CCDP template had an idea in mind.  Perhaps they can expand?  (Don't care much about this point, btw, either will be fine from my perspective.)

ccdps/NNNN-UC-template.md
iawells: > @iawells I'm totally agree with that distinction but I guess what @rannyh is referring is the fact that Kube-Native has not been documented anywhere in this repo and eventually can be used by other documents.

Agreed.

xmulligan:
@vukg the template is fine to be merged, but there are still some comments about the use case your proposed, can you take make that use case a separate PR and we can merge the template

electrocucaracha (approved):

PR #89, opened 2021-03-18
Create a glossary for CNF-WG

### LINKED ISSUE #82: We need a glossary
Common terms likely to be used in multiple use cases or in best practices should have a glossary.  That way, many use cases can do without a glossary because the terms will be globally defined.

jeffsaelens:
This is a first pass to get the basic layout put in place for a glossary. I kept things limited, so we can get the framework in place and then begin to fill in definitions as the group identifies unknown terms.

This PR is to begin work on issue #82

### DIFF
--- a/doc/glossary.md
+++ b/doc/glossary.md
@@ -5,7 +5,7 @@ CNF Working Group Glossary
 Specification
 -------------
 
-### Communinty Definitions
+### Community Definitions
 
 **Cloud-Native** - Cloud native technologies empower organizations to build and run scalable applications in modern, dynamic environments such as public, private, and hybrid clouds. Containers, service meshes, microservices, immutable infrastructure, and declarative APIs exemplify this approach.
 
--- a/doc/glossary.md
+++ b/doc/glossary.md
@@ -7,15 +7,10 @@ Specification
 
 ### Community Definitions
 
-**Cloud-Native** - Cloud native technologies empower organizations to build and run scalable applications in modern, dynamic environments such as public, private, and hybrid clouds. Containers, service meshes, microservices, immutable infrastructure, and declarative APIs exemplify this approach.
+**Cloud-Native** - TBD
+**Cloud-Native Network Function** - TBD
 
-These techniques enable loosely coupled systems that are resilient, manageable, and observable. Combined with robust automation, they allow engineers to make high-impact changes frequently and predictably with minimal toil.
-
-The Cloud Native Computing Foundation seeks to drive adoption of this paradigm by fostering and sustaining an ecosystem of open source, vendor-neutral projects. We democratize state-of-the-art patterns to make these innovations accessible for everyone.
-
-**Cloud-Native Network Function** - A CNF is network functionality delivered in software via cloud native development and delivery practices. This functionality lives within the layers of the OSI Model [9], which is used to define a network's stack. The lower layers (layers 1 and in some cases, layer 2) are provisioned for the higher layers (2-7) to provide transport. These higher layers in this instance act as applications that act upon a network payload (frames, packets datagrams etc). A physical layer 1 networking device should be "flashed" with a complete replacement of its artifacts for updates. The configuration for physical layer 1 is done with an atomic application of a versioned configuration file, which replaces the configuration on the device at once. Virtual Layer 1 (and some layer 2) is managed via templated images and bootstrapping. In contrast, layers 2 through 7 are managed by higher-level orchestration or an established control plane (an orchestrator pushing configuration versus a network protocol modifying a route table).
-
-**Kubernetes** - Kubernetes is a portable, extensible, open-source platform for managing containerized workloads and services, that facilitates both declarative configuration and automation. It has a large, rapidly growing ecosystem. Kubernetes services, support, and tools are widely available.
+**Kubernetes** - Kubernetes is a portable, extensible, open-source platform for managing containerized workloads and services, that facilitates both declarative configuration and automation. It has a large, rapidly growing ecosystem. Kubernetes services, support, and tools are widely available. <Kubernetes.io>
 
 ### CNF-WG Specific Definitions
 
--- a/doc/glossary.md
+++ b/doc/glossary.md
@@ -8,6 +8,7 @@ Specification
 ### Community Definitions
 
 **Cloud-Native** - TBD
+
 **Cloud-Native Network Function** - TBD
 
 **Kubernetes** - Kubernetes is a portable, extensible, open-source platform for managing containerized workloads and services, that facilitates both declarative configuration and automation. It has a large, rapidly growing ecosystem. Kubernetes services, support, and tools are widely available. <Kubernetes.io>
--- a/doc/glossary.md
+++ b/doc/glossary.md
@@ -1,5 +1,5 @@
 
-CNF Working Group Glossary
+Glossary
 ============================
 
 Specification
--- a/doc/glossary.md
+++ b/doc/glossary.md
@@ -1,0 +1,31 @@
+
+CNF Working Group Glossary
+============================
+
+Specification
+-------------
+
+### Communinty Definitions
+
+**Cloud-Native** - Cloud native technologies empower organizations to build and run scalable applications in modern, dynamic environments such as public, private, and hybrid clouds. Containers, service meshes, microservices, immutable infrastructure, and declarative APIs exemplify this approach.
+
+These techniques enable loosely coupled systems that are resilient, manageable, and observable. Combined with robust automation, they allow engineers to make high-impact changes frequently and predictably with minimal toil.
+
+The Cloud Native Computing Foundation seeks to drive adoption of this paradigm by fostering and sustaining an ecosystem of open source, vendor-neutral projects. We democratize state-of-the-art patterns to make these innovations accessible for everyone.
+
+**Cloud-Native Network Function** - A CNF is network functionality delivered in software via cloud native development and delivery practices. This functionality lives within the layers of the OSI Model [9], which is used to define a network's stack. The lower layers (layers 1 and in some cases, layer 2) are provisioned for the higher layers (2-7) to provide transport. These higher layers in this instance act as applications that act upon a network payload (frames, packets datagrams etc). A physical layer 1 networking device should be "flashed" with a complete replacement of its artifacts for updates. The configuration for physical layer 1 is done with an atomic application of a versioned configuration file, which replaces the configuration on the device at once. Virtual Layer 1 (and some layer 2) is managed via templated images and bootstrapping. In contrast, layers 2 through 7 are managed by higher-level orchestration or an established control plane (an orchestrator pushing configuration versus a network protocol modifying a route table).
+
+**Kubernetes** - Kubernetes is a portable, extensible, open-source platform for managing containerized workloads and services, that facilitates both declarative configuration and automation. It has a large, rapidly growing ecosystem. Kubernetes services, support, and tools are widely available.
+
+### CNF-WG Specific Definitions
+
+**Kube-Native** - TBD
+
+**Additional Definitions?**
+
+References
+----------
+
+* CNCF Git [CNCF Cloud Native Definition](https://github.com/cncf/toc/blob/main/DEFINITION.md)
+* Cloud Native Principles [Cloud Native Principles Preamble](https://github.com/cloud-native-principles/cloud-native-principles/blob/master/cloud-native-networking-preamble.md)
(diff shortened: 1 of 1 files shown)


doc/glossary.md
```diff
@@ -0,0 +1,31 @@
+
+CNF Working Group Glossary
+============================
+
+Specification
+-------------
+
+### Communinty Definitions
+
+**Cloud-Native** - Cloud native technologies empower organizations to build and run scalable applications in modern, dynamic environments such as public, private, and hybrid clouds. Containers, service meshes, microservices, immutable infrastructure, and declarative APIs exemplify this approach.
```
iawells: Doesn't actually define what cloud native is, but what it does.  (Per my usual ranting.)  Can we do a little bit better and have the definition before the benefits and the techniques?  Something like 'Software primarily designed to run on a compute cloud and take advantage of functionality specific to running on a cloud.'

doc/glossary.md
```diff
@@ -0,0 +1,31 @@
...
+
+**Cloud-Native Network Function** - A CNF is network functionality delivered in software via cloud native development and delivery practices. This functionality lives within the layers of the OSI Model [9], which is used to define a network's stack. The lower layers (layers 1 and in some cases, layer 2) are provisioned for the higher layers (2-7) to provide transport. These higher layers in this instance act as applications that act upon a network payload (frames, packets datagrams etc). A physical layer 1 networking device should be "flashed" with a complete replacement of its artifacts for updates. The configuration for physical layer 1 is done with an atomic application of a versioned configuration file, which replaces the configuration on the device at once. Virtual Layer 1 (and some layer 2) is managed via templated images and bootstrapping. In contrast, layers 2 through 7 are managed by higher-level orchestration or an established control plane (an orchestrator pushing configuration versus a network protocol modifying a route table).
```
iawells: Define 'network function' first?

iawells: Anything that talks networking lives in the OSI model (it includes the application layer), so that's a tautology.  You have more specific information, but even that is a bit wonky; we would consider BGP a network function.

Proposal: 

Network functions [NB: separating this from CNF, so that we could define VNF or PNF if we had to] consist of components whose primary task is to enable the network to function and to control its behaviour.  This might include functionality transforming packets, such as network encaps, switching, routing, firewalls or NAT; information sharing systems propagating network information, such as BGP, DHCP or DNS; or systems for gathering network information, including SNMP and BGP-LS.  More complex network functions, such a mobile packet cores, may combine multiple behaviours.

Software network functions are a specific type of application with a focus on the behaviour of the network.

doc/glossary.md
```diff
@@ -0,0 +1,31 @@
...
+
+**Kubernetes** - Kubernetes is a portable, extensible, open-source platform for managing containerized workloads and services, that facilitates both declarative configuration and automation. It has a large, rapidly growing ecosystem. Kubernetes services, support, and tools are widely available.
+
+### CNF-WG Specific Definitions
+
+**Kube-Native** - TBD
```
iawells, line 18: A subset of cloud-native applications specifically designed to run ion Kubernetes and built to take advantage of its inbuilt features and functionality.

iawells (changes requested):

doc/glossary.md
jeffsaelens: I'm against redefining things over and over again. If we want it called out, we should link to something as network functionality is not a novel concept here.

iawells: Well, you're attempting to define it here, and I'm not sure I have seen a good definition of it, so I actually do appreciate your attempt.

CNF brings 'network function' and 'cloud native' together, though, and would be a shorter definition and more comprehensible if it did refer out to both concepts.

jeffsaelens: I don't think we need to define network functions themselves. Also, I hesitate to include a big list of features/protocols as they then require to be defined themselves. I agree that anything talking networking lives in the OSI model, but I don't see what is wrong with saying the software we are describing lives in the OSI model vs providing a bunch of references/mechanisms to how data is treated at the various layers of said model.

We can debate/change the definition above, but I don't think we should spend a lot of time defining things that aren't unique to the space we are exploring.

doc/glossary.md
jeffsaelens: The principles repo, vendors docs and other credible sources provide concrete definitions we could pull from. I just fork-lifted this because for good or ill its the official definition provided by the CNCF and we are working under their umbrella.

doc/glossary.md
```diff
@@ -0,0 +1,31 @@
...
+============================
+
+Specification
+-------------
+
+### Communinty Definitions
+
+**Cloud-Native** - Cloud native technologies empower organizations to build and run scalable applications in modern, dynamic environments such as public, private, and hybrid clouds. Containers, service meshes, microservices, immutable infrastructure, and declarative APIs exemplify this approach.
+
+These techniques enable loosely coupled systems that are resilient, manageable, and observable. Combined with robust automation, they allow engineers to make high-impact changes frequently and predictably with minimal toil.
```
electrocucaracha: is this *robust automation* referring to the CI processes? if that is the case we can make it more explicit like:

```suggestion
These techniques enable loosely coupled systems that are resilient, manageable, and observable. Combined with Continuous Integration processes, engineers can make high-impact changes frequently and predictably with minimal toil.
```

doc/glossary.md
electrocucaracha: Just move the place of the acronym 
```suggestion
**Cloud-Native Network Function (CNF)** - Delivers network functionality in software via cloud native development and delivery practices. This functionality lives within the layers of the OSI Model [9], which is used to define a network's stack. The lower layers (layers 1 and in some cases, layer 2) are provisioned for the higher layers (2-7) to provide transport. These higher layers in this instance act as applications that act upon a network payload (frames, packets datagrams etc). A physical layer 1 networking device should be "flashed" with a complete replacement of its artifacts for updates. The configuration for physical layer 1 is done with an atomic application of a versioned configuration file, which replaces the configuration on the device at once. Virtual Layer 1 (and some layer 2) is managed via templated images and bootstrapping. In contrast, layers 2 through 7 are managed by higher-level orchestration or an established control plane (an orchestrator pushing configuration versus a network protocol modifying a route table).
```

doc/glossary.md
electrocucaracha: Technically, we can have *docker-swarm-native* and *mesos-native* applications right?

doc/glossary.md
```diff
@@ -0,0 +1,31 @@
+
+CNF Working Group Glossary
+============================
+
+Specification
+-------------
+
+### Communinty Definitions
```
electrocucaracha:
```suggestion
### Community Definitions
```

doc/glossary.md
```diff
@@ -0,0 +1,31 @@
+
+CNF Working Group Glossary
```
electrocucaracha: I think that we can assume the context of this document
```suggestion
Glossary
```

doc/glossary.md
iawells: I think as things stand this is a C&P from the CNCF definition. If we are just going to quote it (and that seems to be somewhat at odds with @jeffsaelens' other comments we should reference things) we should not change it.

doc/glossary.md
iawells: We can have AWS-native and OpenStack-native.  Clouds are not 'things that run containers'.

doc/glossary.md
jeffsaelens: Everything was a quote. I pulled all three defs from the links down in the references. I'm down to modify these, but I just put them as a starting point mainly to get a template in place for us to submit PRs against to add terms. I intentionally did not include a ton of definitions in the first PR because this will be the 4th-5th time I've been in debates around what the majority of these terms mean. (without almost the same crowd of people)

I figured, just get the glossary place holder in the repo so that the battle of PRs could begin a new. I strongly suggest that going forward we limit a PR to a single term. That way terms not caught in philosophical debates can be merged quickly.

doc/glossary.md
CsatariGergely: If we stick to the CNCF definition we should also put a reference to it. 

I agree with @iawells , that the current CNCF cloud native definition does not define what cloud native is...

doc/glossary.md
CsatariGergely: I like more the [CNCF TUG's 1st CNF definition](https://docs.google.com/document/d/1-zqxz5bdCLTuOEvi2ybADR3PcmzbBhNt6YkNnvx-KoA/edit#) "A Cloud Native Virtual Network Function - or CN F for short - is an implementation of a Virtual Network Function (as defined by ETSI GS NFV 003 V1.4.1 (2018-08) that adheres to the CNCF Cloud Native Definition. "
or maybe the [CNCF TUG's 2nd CNF definition](https://github.com/cncf/telecom-user-group/raw/master/whitepaper/CNCF%20_TUG_Cloud%20Native%20Thinking%20for%20Telecommunications.pdf) "A cloud native network function (CNF) is a cloud native application that implements network functionality. A CNF consists of one or more microservices and has been developed using Cloud Native Principles including immutable infrastructure, declarative APIs, and a “repeatable deployment process”.
An example of a simple CNF is a packet filter that implements a single piece of network functionality as a microservice. A firewall is an example of a CNF which may be composed of more than one microservice (i.e. encryption, decryption, access lists, packet inspection, etc.).
The CNFs themselves, by nature of following cloud native principles, are also composable and can implement and facilitate more complex network functionality if needed.
CNFs and cloud native principles can be used when implementing applications and network infrastructure which needs to meet standards such as 3GPP. For example, an Evolved Packet Core’s (EPC) Serving Gateway could be implemented as a CNF and support all 3GPP requirements (eg. S1 protocol stack) allowing integration with other EPC services."

I think we should not redefine network function as we clearly refer to VNF-s with the CNF definition and in ETSI NFV NF already have a definition.

fkautz: We'd also like to have CNFs target enterprise organizations, which won't adhere to ETSI. I think it's ok to look there for inspiration, but we probably shouldn't tie it to "requires ETSI" or we'll lose that crowd.

doc/glossary.md
fkautz: Correct, but for the initial scope of this WG, I believe we're focusing on kube-native and will expand out later.

doc/glossary.md
jeffsaelens: References are in the bottom, but we could add citations to ensure the link is made for the reader.

doc/glossary.md
jeffsaelens: I actually like the "native landscape". Cloud Native has morphed into an impossible to reach, but always aspired to ideal in mind. Even the original thought of "it runs natively in a cloud and takes advantage of the services offered" was a loaded concept as no two clouds like the same. Specifying what the target platform is, while also striving to meet the holy grail that is cloud nativeness seems like a practical approach (though how tedious it is to scale is yet to be seen).

jeffsaelens (commented):
I'm willing to scrap definitions and just change this to a list of place holders if needed. That way we could at least get the template in place.

doc/glossary.md
iawells: I don't want to go down the well-trodden path that we define CNFs by how they're implemented. We define them by what they do and get people who implement them to suggest what works best for their implementation. Our best practices are a means to record and share that.

A network function is a thing that does network stuff. A cloud native network function is one that does it while it runs on a cloud and has been designed to benefit from running on a cloud.

Its use of microservices should very likely be recommended by us as a best practice for CNF dev teams to implement (until someone comes up with a better design) because it has a lot of benefit when you run on the cloud, and that makes it a Best Practice. But that's not what makes it either a network function or cloud native, and so it doesn't belong in the glossary.

xmulligan (approved):

taylor (approved):
lgtm as a start

doc/glossary.md
```diff
@@ -0,0 +1,27 @@
...
+Specification
+-------------
+
+### Community Definitions
+
+**Cloud-Native** - TBD
+
+**Cloud-Native Network Function** - TBD
+
+**Kubernetes** - Kubernetes is a portable, extensible, open-source platform for managing containerized workloads and services, that facilitates both declarative configuration and automation. It has a large, rapidly growing ecosystem. Kubernetes services, support, and tools are widely available. <Kubernetes.io>
```
taylor, line 14:
```suggestion
**Kubernetes** - Kubernetes is a portable, extensible, open-source platform for managing containerized workloads and services, that facilitates both declarative configuration and automation. It has a large, rapidly growing ecosystem. Kubernetes services, support, and tools are widely available. ref: https://kubernetes.io/docs/concepts/overview/what-is-kubernetes/
```

taylor (changes requested):

taylor (approved):
lgtm as start

PR #98, opened 2021-03-22
Clarifying the responsibilities of the Governing Chairs

taylor:
The Chair role responsibilities left some things ambiguous and was overreaching with a couple of items.  This role is focused on governing of the working group. Other activities should be left to the community to decide and discuss.

As seen in The Art of Community by Jono Bacon (2009), common responsibilities for a council (aka governing body) include:
- membership approval/rejection
- conflict resolution
- project values
- community process changes
- ordaining governance bodies
- direction

The governing council does not
- act as a bottle neck to choose people for small roles
- set feature direction when there are excellent contributors not on the council
- approve formation of a new team

Reference:
- https://www.jonobacon.com/books/artofcommunity/ ([Creative Commons licensed PDF](https://drive.google.com/file/d/1EI6YcKlTdzojLD4RdVjYVlmFRTNzzge0/view))

PR #99, opened 2021-03-23
Add rules of arbitration for representation limits

### LINKED ISSUE #83: Update charter with org-merge voting rules
From the discussion, the text we like most is:

> As an aside, note that this is more about the reputation of this group being co-opted in the future than it is about it being seen as a mouthpiece for a vested interest in the present. If it's obviously a mouthpiece, that's an issue, but it will impact our reputation and therefore adoption of our recommendations. If our group is co-opted in the future, however, then a group will be able to take advantage of our established reputation.
> 
> 1. Define representation groups that we wish to control representation on. For instance: one chair per independent organisation, 30% of community votes, 30% of tech leads (to avoid majorities of conspirators).
> 2. A organisation is defined as a single company.
> 3. If a member (define 'member') raises a concern of independence: first they must demonstrate a relationship between organisations that lead to the conclusion they should be treated as a single organisation (e.g ownership, contractor-contractee); second, the relationship will be debated, since this does not necessarily demonstrate that the organisations are acting together; and third, action will be taken only if a majority of uninvolved members (choose your membership set and quorum here) votes for it. The action would be that the multiple entities would be considered one organisation under the charter.
> 4. If, through this process, an organisation becomes overrepresented, the seats for that organisation will all be put up for a special election. The seats are open to all members, both previous holders and any other nominees.
> 5. In any elections (presuming we use STV or something similar), the members with least votes in an overrepresented group will be excluded from election (this should already be in some other part of the charter). Where there are multiple candidates with the same number of least votes, it's a coin toss (ditto).
> 6. The decision on independence or dependence may be revisited after a period of not less than three months. A member may raise it for discussion and the process in point 3 shall be repeated. No immediate re-election of members shall occur if the organisations now are deemed independent; this will be taken into consideration in the next natural election cycle.
> To game the above rules, you'd require a majority of the membership conspiring against some part of the community that had overrepresentation if and only if it were considered a bloc, and that bloc would need some corporate relationship, so it would be an unproductive thing to try and do unless a real concern existed. (I can't claim 'everyone not in my company is conspiring against me', for instance, because I haven't got valid grounds to raise the concern.)
> 
> It does not present multiple independent blocs agreeing to vote in concert. What prevents that is a healthy and diverse membership, and there's nothing the rules can do about that.
>

iawells:
The first change in this pull request also includes some cosmetics on the wording of the charter.

NB: I have not yet changed the 'one company, one vote' rule, since we're about to start an election cycle.

### DIFF
--- a/charter.md
+++ b/charter.md
@@ -90,7 +90,7 @@ The CNF WG will be co-chaired by 3 people - one co-chair from the Kubernetes (K8
 
 **Representation**
 
-A maximum of one person from any one company/organization at any time may hold a chair role.
+A maximum of one person from any one entity may hold a chair role at any given moment.  See the [Independence](#independence) section for further details.
 
 #### Tech Lead
 
@@ -102,9 +102,7 @@ Tech leads are primarily focused on the technical execution of the working group
 
 **Representation**
 
-A maximum of one-third representation in tech lead roles is permitted from any one company/organization at any time. If the results of an election result in greater than 1/3 representation, the lowest vote getters from any particular company/organization will be removed until representation is less than one-third.
-
-If percentages shift because of job changes, acquisitions, or other events, sufficient tech leads must resign until max one-third representation is achieved. If it is impossible to find sufficient teach leads to resign, the entire company/organization’s representation will be removed and new special elections held. In the event of a question of company/organization membership (for example evaluating independence of corporate subsidiaries) a majority of all non-involved members will decide.
+A maximum of one-third of the tech lead roles may be held by people from any one entity at any given moment.  See the [Independence](#independence) section for further details.
 
 
 #### Everyone else (eg. other members)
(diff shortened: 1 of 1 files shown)


rabi-abdel (approved):

xmulligan (approved):

CsatariGergely (approved):

iawells:
Merged with the changes in the original since the original pull request.

charter.md
```diff
@@ -132,4 +132,44 @@ For the purpose of this document, “Company” shall mean any entity (Company-A
...
+
+The CNF WG strives to maintain its editorial independence from the companies and organisations that provide its members.  To this end, this charter sets limits on the involvement any given entity might have in its operation, in order to maintain the integrity of its output as a consensus view of industry technical experts.
+
+To this end, representation from a single entity for a given role is limited, to avoid undue influence.  This section determines how companies, organisations and individual interested parties are determined to be independent of each other, and the way in which this affects elections.
+
+#### Determination of independence
+
+1. An entity consists of a single company or organisation, a related group of companies or organisations in common ownership, or an individual acting independently of their employment.
+
+2.  Entities may potentially not be independent if:
```
electrocucaracha: nit: extra space
```suggestion
2. Entities may potentially not be independent if:
```

electrocucaracha (approved):

charter.md
```diff
@@ -78,31 +78,31 @@ Deliverables
...
 #### Chairs
 
-The CNF WG will be co-chaired by 3 people: at least one co-chair from the Kubernetes (K8s) Community, one from the Service Provider (SP) Community and one from the CNF Developer (Dev) Community to have equal representation.
-- *Primary role of Chairs is governance of the group.*
-- The Chairs are responsible for ensuring that group meetings are facilitated effectively, while also engaging group members in leadership roles. Effective facilitation includes the following activities:
+The CNF WG will be co-chaired by 3 people - one co-chair from the Kubernetes (K8s) Community, one from the Service Provider (SP) Community and one from the CNF Developer (Dev) Community - in order to maintain equal representation.
+
+- *The primary role of Chairs is to run operations and the governance of the group.*
+- The Chairs are responsible for ensuring that group meetings are planned and facilitated effectively, while also engaging group members in leadership roles. Effective facilitation includes (but is not limited to) the following activities:
+    - setting the agenda for meetings
```
taylor:
```suggestion
- The Chairs are responsible for ensuring that group meetings are planned and facilitated effectively, while also engaging group members in leadership roles. Effective facilitation includes (but is not limited to) the following activities:
```

Remove "setting the agenda for meetings"

charter.md
```diff
@@ -132,4 +132,44 @@ For the purpose of this document, “Company” shall mean any entity (Company-A
...
+
+3. If a member raises a concern of that two entities are not independent, or, conversely, that a single entity should be considered as two separate entities, the member shall make their case and the co-chairs will consider it.
+
+4. If they deem the case valid, the relationship and its effects will be discussed by members.  The existence of a relationship does not necessarily demonstrate that the organisations are acting in concert.
+
+5. A vote of all uninvolved members will be held to determine whether or not the multiple parties should be considered one entity under the rules of this charter.
+
+#### Reconsidering the decision
+
+The decision on independence or dependence of entities may be revisited after a period of not less than three months.  After this time, any member may raise it for discussion via the above process.
```
taylor, line 162: Where did the three month time limit come from?

taylor (changes requested):

charter.md
taylor:
```suggestion
- *Primary role of Chairs is governance of the group.*
- The Chairs are responsible for ensuring that group meetings are facilitated effectively, while also engaging group members in leadership roles. Effective facilitation includes the following activities:
```

taylor (changes requested):

charter.md
taylor: Update from https://github.com/cncf/cnf-wg/pull/98

iawells: why?

charter.md
iawells: Arbitrary choice.

charter.md
iawells: ok, done in the other proposed change

taylor (approved):

PR #101, opened 2021-03-23
Proposing to add myself to the interested parties

nickolaev:


taylor (approved):
Party with no company affiliation (eg. independent).
Company they work for is not listed as an interested party.

PR #122, opened 2021-04-01
add co-chairs

xmulligan:


README.md
```diff
@@ -12,6 +12,11 @@ The CNF Conformance program will evaluate cloud nativeness based on the set of b
 
 ## Meetings
 
+### Current Co-Chairs
+Service Provider Co-Chair: Jeffrey Saelens
```
lixuna: Do we want to add bullet points?

* Service Provider Co-Chair: Jeffrey Saelens
* CNF Developer Co-Chair: Ian Wells
* Kubernetes Co-Chair: Taylor Carpenter

jeffsaelens (approved):

README.md
```diff
@@ -12,6 +12,11 @@ The CNF Conformance program will evaluate cloud nativeness based on the set of b
...
+CNF Developer Co-Chair: Ian Wells
+Kubernetes Co-Chair: Taylor Carpenter
```
xmulligan:
```suggestion
* Service Provider Co-Chair: Jeffrey Saelens
* CNF Developer Co-Chair: Ian Wells
* Kubernetes Co-Chair: Taylor Carpenter
```

README.md
xmulligan: yes, good point :)

iawells:
I am not going to object to doing it this way, but I wonder if adding this in the charter makes sense (we're not changing the charter when the co-chairs change), and it would additionally mean we can't be inconsistent if it's only in one place.

xmulligan:
@iawells I agree. I think the Governance should be pulled out of the charter as a separate file similar to the contributors.md. That would also solve your issue

iawells:
If anyone figures out how to use include files in markdown, our lives will be easier (though unfortunately I don't think this is a feature of github's markdown engine)

taylor:
@iawells I agree with governance being pulled out of the charter. Are you okay with addressing that in a new PR?

fkautz (approved):

electrocucaracha (approved):

PR #168, opened 2021-05-10
add 5G RAN use cases

ashishsanjaysharma:
Use cases:
- UC-0001: Synchronization/Timing in Telco CNF world
   - discussion https://github.com/cncf/cnf-wg/discussions/174
- UC-0002: Support of Multi-Communication Interface in Telco CNF world

### DIFF
--- a/use-case/5G_RAN/0001-UC-sync-timing-ran.md
+++ b/use-case/5G_RAN/0001-UC-sync-timing-ran.md
@@ -1,0 +1,107 @@
+# UC-0001: Synchronization/Timing in Telco CNF world
+
+  
+
+## Glossary
+
+> RAN - Radio Access Network
+> DU - Distributed Unit support RLC, MAC, Physical layers
+> CU - Centralized Unit support SDAP, RRC, PDCP protocol layers
+> RU - Radio Unit
+> gNB - gNodeB
+> PTP - Precision Time Protocol
+
+  
+
+## Involved processes
+
+- [x] Development
+
+- [x] Deployment
+
+- [x] System integration
+
+- [x] Network integration
+
+- [ ] Lifecycle management
+
+- [x] Operations
+
+  
+
+## Involved actors / personas
+
+  
+
+> Business customer of CSP
+Operations Team of CSP
+CNF Vendor
+Cloud provider
+Datacenter Operator/provider
+
+  
+
+## Involved system entities
+
+  
+
+>Data Center Network
+Wide Area Network
+Cloud Native Platform
+CNF X
+RAN
+MANO
+
+  
+
+## Situation
+
+> 3GPP 5G gNB have a flexible architecture to support a wide range of deployments. The gNB can be split into CU, DU, and RU. The CU provides the support of a higher layer protocol stack(Layer 3 and above). The DU runs a lower-layer protocol stack(L1, L2), and RU does the radio and Low-PHY processing. The DU requires real-time performance and strict timing or synchronization to perform all the Layer 1, Layer 2 processing and scheduling.
+
+> To meet 5G requirements and processing demands, the network nodes must be synchronized. There are different ways to distribute the timing, like through GPS or any master PTP clock in the network with IEEE 1588v2 PTP.
+
+> In a non-virtualized network(DU/CU), the entire monolithic stack runs directly on ASIC or hardware at line rates, which aids in maintaining the synchronization/timing.
+
+> In container-based virtualized systems(vDU/vCU) or CNFs, the platform manages the workloads. In cloud native design, a monolithic vDU or vCU would result in many individual workloads. 5G has critical timing/synchronization requirements because of new services, new architectures and new technologies[1]. A time sync of ~1.5 microseconds or less is required to support the new spectrum and RAN features/applications[2]. The workloads need to maintain tight synchronization to meet such requirements.
+
+> The nature of the telco workload is compute-network intensive, demanding real-time performance. The orchestration platforms need to support CNFs to meet their timing synchronization requirements.
+
+  
+
+## Expected behaviour
+
+  
+
+> The cloud native platform should be able to maintain the synchronization/timing of the telco workloads.
+
+  
+
+## Challenges and limitations with Kube-native approach
+
+  
+
+> The K8s egress/ingress gateway is limited to web-based traffic. They are not aware of time stamped packets, like PTP packets. These packets are time-critical and they do not identify or process differently. The existing gateways might introduce a delay for many unknown reasons during the distribution of synchronization/timing packets to workloads. This unwanted delay may lead the workloads to lose their synchronization. The IEEE 1588v2 PTP have taken care of such scenarios of loss of synchronization, but the container platform may lack the capabilities internally that can enable the workloads to manage their synchronization/timing.
+
+  
+
+### Established practices to overcome challenges and limitations
+
+  
+
+> To be developed
+
+  
+
+### What needs to be done differently in order to overcome challenges and limitations
+
+  
+
+> Kubernetes could have a time-sensitive gateway or a particular gateway that keeps track of synchronization/timing distributed from the network or GPS to workloads. This would help workload maintain their synchronization across the network. There could also be a correction mechanism internally in the platform to maintain the synchronization/timing.
+
+  
+
+### References
+
+> [1] [https://wsts.atis.org/wp-content/uploads/sites/9/2018/11/2-01_China-Mobile_Li_5G-Synchronization_v5.pdf](https://wsts.atis.org/wp-content/uploads/sites/9/2018/11/2-01_China-Mobile_Li_5G-Synchronization_v5.pdf)
+
+> [2] [https://www.ericsson.com/en/blog/2019/8/what-you-need-to-know-about-timing-and-sync-in-5g-transport-networks](https://www.ericsson.com/en/blog/2019/8/what-you-need-to-know-about-timing-and-sync-in-5g-transport-networks)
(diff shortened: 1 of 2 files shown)


use-case/5G_RAN/0001-UC-sync-timing-ran.md
```diff
@@ -0,0 +1,107 @@
...
+
+> The nature of the telco workload is compute-network intensive, demanding real-time performance. The orchestration platforms need to support CNFs to meet their timing synchronization requirements.
+
+  
+
+## Expected behaviour
+
+  
+
+> The cloud native platform should be able to maintain the synchronization/timing of the telco workloads.
```
iawells, line 70: While this statement is true, we need a threshold to test against, I think.  There doesn't seem to be a universal opinion of what a platform should provide.  RAN vendors express themselves in different ways, such 'you will put this content in this platform config file' (one option, possibly of many) and 'if the application passes its system tests you have it right' (true, but that testing is very late in the process and doesn't allow the platform to be independent of the application).  We actually need 'the application runs if the platform *behaves* in this way' if we want the application to be properly independent of the platform. 

Also, there are two different elements of what's going on here and they're being mixed up.  One is how the PTPSyncE clock is provided to the application, which is necessary for the application to run.  The other is the real-time behaviour of the platform, and please note that there is no detailed specification of what real-time means - what it means to a UI developer and to someone doing hardware control is very different, and RAN falls somewhere between those two.

This use case is very good, but we need to add some details to it to make it testable, and we will probably have to add more in the future (let's not assume we must have it perfectly correct first time or we'll never commit anything).

use-case/5G_RAN/0001-UC-sync-timing-ran.md
```diff
@@ -0,0 +1,107 @@
...
+
+> To be developed
+
+  
+
+### What needs to be done differently in order to overcome challenges and limitations
+
+  
+
+> Kubernetes could have a time-sensitive gateway or a particular gateway that keeps track of synchronization/timing distributed from the network or GPS to workloads. This would help workload maintain their synchronization across the network. There could also be a correction mechanism internally in the platform to maintain the synchronization/timing.
```
fkautz, line 94: Should this be in kubernetes? Why not drive this as a requirement from the host? Are you expecting the workloads to process the PTP packet, or is the kernel dealing with this? It may be better to expose this through the clock_gettime syscall if the syscall overhead isn't too high.

Also, if accuracy is paramount, we might need to consider CPU pinning here since the clock for each core is independent. Hopefully we don't need to do this though.

use-case/5G_RAN/0001-UC-sync-timing-ran.md
```diff
@@ -0,0 +1,107 @@
...
+
+> To meet 5G requirements and processing demands, the network nodes must be synchronized. There are different ways to distribute the timing, like through GPS or any master PTP clock in the network with IEEE 1588v2 PTP.
+
+> In a non-virtualized network(DU/CU), the entire monolithic stack runs directly on ASIC or hardware at line rates, which aids in maintaining the synchronization/timing.
+
+> In container-based virtualized systems(vDU/vCU) or CNFs, the platform manages the workloads. In cloud native design, a monolithic vDU or vCU would result in many individual workloads. 5G has critical timing/synchronization requirements because of new services, new architectures and new technologies[1]. A time sync of ~1.5 microseconds or less is required to support the new spectrum and RAN features/applications[2]. The workloads need to maintain tight synchronization to meet such requirements.
```
fkautz, line 60: In a Linux container, the clock is likely not virtualized. We should confirm though.

fkautz (changes requested):

use-case/5G_RAN/0001-UC-sync-timing-ran.md
iawells: You need accurate time sync; I'm not actually clear if clock_gettime is what is used to retrieve the clock or if there's more to it, since there's no concrete guarantee on how long a Linux syscall takes to return when you call it.  A RAN vendor would be welcome here to speak on this with more certainty.  I understand that PTP is generally processed by the host (and therefore the platform) and not passed to the app as raw packets but as deduced timing information.

You also, independently (though for similar reasons), need RT behaviour down to a high level of precision - specifically, delivering a specific quantity of CPU to a vDU within a given time period (in the msec frame).  This is not an easy thing to guarantee with conventional pre-emptive multitasking systems.   CPU pinning is very likely to be required to help with that, but there's more than that to it.

CPU pinning is in fact not a guarantee of anything at the application level; it's a guarantee that certain things won't happen in the system, with an _implication_ that performance will be better as a result (but no promises as to the achieved performance level).  NUMA is another feature of the same type.  Alone, they don't get you to the guarantee mentioned above.

The key distinction here is between 'guarantee' and 'best effort'.  Best effort is not enough for some of the work running here.  Statistically analysed best effort (e.g. '90% of the time you get this') might be acceptable, but it's hard to prove that's what you're getting, because you're proving a negative - specifically it's hard to demonstrate that there aren't rare cases (once a week, say) where that guarantee doesn't hold.  Things like web apps are fine with best effort, because what they're effectively asking for a weak statistical guarantee ('don't delay a web response by 100ms too often') and it's so all-encompassing that exceptions are vanishingly rare and can be addressed one by one by the ops team - so this is where our workloads differ from the norm.

use-case/5G_RAN/0001-UC-sync-timing-ran.md
```diff
@@ -0,0 +1,107 @@
...
+  
+
+## Glossary
+
+> RAN - Radio Access Network
+> DU - Distributed Unit support RLC, MAC, Physical layers
+> CU - Centralized Unit support SDAP, RRC, PDCP protocol layers
+> RU - Radio Unit
+> gNB - gNodeB
+> PTP - Precision Time Protocol
```
sishbi:
```suggestion
- PTP - Precision Time Protocol
```
This list should use a bullet '-' instead of the '>' symbol as the list doesn't render correctly otherwise.

use-case/5G_RAN/0002-UC-multi-interface.md
```diff
@@ -0,0 +1,65 @@
+# UC-0002: Support of Multi-Communication Interface in Telco CNF world 
+
+## Glossary
+
+> RAN - Radio Access Network
+> DU - Distributed Unit  support  RLC, MAC, Physical layers
```
sishbi, line 6: Needs an explanation of what 'RLC' and 'MAC' mean

sishbi (changes requested):
I don't fully understand the RAN terminology but I can read along.
The formatting of these 2 use cases needs fixing to remove the additional quoted text marker: '>'
Also the lists should use '-' to create a bulleted list.

This has an interesting proposal for using a multi-network interface CNI that enables cross-Pod IPC communication

use-case/5G_RAN/0001-UC-sync-timing-ran.md
```diff
@@ -0,0 +1,107 @@
...
+
+- [x] Deployment
+
+- [x] System integration
+
+- [x] Network integration
+
+- [ ] Lifecycle management
+
+- [x] Operations
```
electrocucaracha:
```suggestion
- [x] Development
- [x] Deployment
- [x] System integration
- [x] Network integration
- [ ] Lifecycle management
- [x] Operations
```

use-case/5G_RAN/0001-UC-sync-timing-ran.md
electrocucaracha: @sishbi suggestion applies to the rest of the document

use-case/5G_RAN/0001-UC-sync-timing-ran.md
```diff
@@ -0,0 +1,107 @@
...
+
+> The cloud native platform should be able to maintain the synchronization/timing of the telco workloads.
+
+  
+
+## Challenges and limitations with Kube-native approach
+
+  
+
+> The K8s egress/ingress gateway is limited to web-based traffic. They are not aware of time stamped packets, like PTP packets. These packets are time-critical and they do not identify or process differently. The existing gateways might introduce a delay for many unknown reasons during the distribution of synchronization/timing packets to workloads. This unwanted delay may lead the workloads to lose their synchronization. The IEEE 1588v2 PTP have taken care of such scenarios of loss of synchronization, but the container platform may lack the capabilities internally that can enable the workloads to manage their synchronization/timing.
```
electrocucaracha, line 78: I just wondering if [Service Topology awareness](https://kubernetes.io/docs/concepts/services-networking/service-topology/) can help on this Kube-Native approach

taylor:
Discussion for timing on K8s and linux hosts -> https://github.com/cncf/cnf-wg/discussions/174

use-case/5G_RAN/0001-UC-sync-timing-ran.md
iawells: What would a 'virtualised' clock mean, exactly?  I'm unclear on your point.

pgoyal01: "In cloud native design, a monolithic vDU or vCU would result in many individual workloads" is the intent here to say that a redesign of a monolithic application will result in multiple components? If that is the case, let us say so clearly.

taylor:
We are merging this PR into a new branch, https://github.com/cncf/cnf-wg/tree/wip-5g_ran_use_cases, where we will continue to work on items outlined in the the PR comments and Monday CNF WG meeting.  We may create a PR for each use case when adding to main.

PR #171, opened 2021-05-13
Update glossary.md

tliron:
Lots of new terms and some streamlining of existing terms.


doc/glossary.md
```diff
@@ -4,18 +4,34 @@
 
 ### Community Definitions
 
-**Cloud Native** - TBD
+**Cloud-Native**: TBD
 
-**Cloud Native Network Function** - A cloud native network function (CNF) is a [cloud native application](https://github.com/cncf/glossary/blob/main/definitions/cloud_native_apps.md) that provides network functionality, which is developed using cloud native principles. 
+**Cloud-Native Network Function (CNF)**: A network function that is a [cloud-native application](https://github.com/cncf/glossary/blob/main/definitions/cloud_native_apps.md), i.e. developed using cloud-native principles.
```
iawells: Per the extensive discussion in the channel, cloud-native network function and CNF are not synonyms, at least not to everyone.  I think we should basically define the 'CNF' acronym simply to say 'has variable meanings' and caution against using it without qualification.

doc/glossary.md
```diff
@@ -4,18 +4,34 @@
...
 
-**Cloud Native** - TBD
+**Cloud-Native**: TBD
 
-**Cloud Native Network Function** - A cloud native network function (CNF) is a [cloud native application](https://github.com/cncf/glossary/blob/main/definitions/cloud_native_apps.md) that provides network functionality, which is developed using cloud native principles. 
+**Cloud-Native Network Function (CNF)**: A network function that is a [cloud-native application](https://github.com/cncf/glossary/blob/main/definitions/cloud_native_apps.md), i.e. developed using cloud-native principles. 
+
+**Kubernetes**: [Kubernetes.io](https://kubernetes.io/) is a portable, extensible, open-source platform for managing workloads and services, that facilitates both declarative configuration and automation. It has a large, rapidly growing ecosystem. Kubernetes services, support, and tools are widely available.
+
+**Network function**: An application that provides network functionality. Network functions may be components of network services.
```
iawells: Physical network functions are not applications, and PNF is an occasionally used term.

doc/glossary.md
```diff
@@ -4,18 +4,34 @@
...
+
+**Network service**: TBD
```
iawells, line 16: I wonder if we'd be best avoiding the use of this term.

doc/glossary.md
```diff
@@ -4,18 +4,34 @@
...
+
+**Physical Network Function (PNF)**: A network function distinguished by its encapsulation and isolation in discrete hardware. Internally it may use various technologies, including virtualization, containerization, cloud platforms, etc.
```
iawells, line 18: Most PNF do not advertise how much software is used in their construction, and they're often (if not always) quite hardware-centric.  Either way, I could define it by what it looks like from the outside, not how it's put together; to the user, it's basically a sealed box with firmware and standardised behaviour.

doc/glossary.md
```diff
@@ -4,18 +4,34 @@
...
+
+**Virtualized Network Function (VNF)**: A network function deployed entirely or primarily using virtualization technologies, which comprise both hardware and software features and include CPU, RAM, and GPU support, as well as bus features such as SR-IOV. Virtualization technologies are available in hypervised virtual machines as well as containers (see: Containerized network function). Note that a VNF is not necessarily a cloudified network function. Indeed, the history of NFV—the effort to virtualize network functions—predates the move to cloud platforms.
```
iawells, line 20: See channel discussion.  This tells me a lot of things, but I still don't know if I would use the term 'VNF' to include CNFs or not.

doc/glossary.md
```diff
@@ -4,18 +4,34 @@
...
 
-**Kubernetes** - Kubernetes is a portable, extensible, open-source platform for managing containerized workloads and services, that facilitates both declarative configuration and automation. It has a large, rapidly growing ecosystem. Kubernetes services, support, and tools are widely available. [Kubernetes.io](https://kubernetes.io/)
 
 ### CNF WG Specific Definitions
 
-**Kube-Native** - TBD
+**Cloudified network function**: A network function that designed to be deployed on cloud platforms, including Kubernetes (see: Kubernetes Network Function). A cloudified network function can use virtualization features (in which case it would also be considered to be a VNF) as well as containers (in which case it is also be considered to be a containerized network function). A cloudified network function is not necessarily a Cloud-native Network Function (CNF). To be classified as such it must adhere to cloud-native design principles.
```
iawells: Ooh, that's a subtle distinction.

doc/glossary.md
```diff
@@ -4,18 +4,34 @@
...
+
+**Containerized network function**: A network function deployed entirely or primarily as one or more containers. Should not be referred to by the "CNF" acronym in order to avoid confusion with Cloud-native Network Function (CNF). Note that a containerized network function is not necessarily a Kubernetes Network Function (KNF) as it is possible to use containers in other and simpler platforms. Thus, it is also not necessarily a cloudified network function. Also note that some containerization technologies include virtualization features (see also: VNF).
```
iawells: Again: if I polled an audience of people about what CNF meant, the majority would _not_ say 'cloud native network function'.

doc/glossary.md
```diff
@@ -4,18 +4,34 @@
...
+
+**Kubernetes Network Function (KNF)**: A cloudified network function deployed on Kubernetes. Note that a KNF is not necessarily a containerized network function as there are ways to employ virtualization features in containers running in Kubernetes, and indeed to run full virtual machines.
```
iawells: Hmmm.  You're not wrong, but is this term going to be useful to us?

iawells (commented):
Another way of looking at this: a network function can be one or more of:

- containerised
- demonstrating adherence to cloud native design principles: detectable iff you look at its code and internal design, which are not always visible to users; dependent upon which set of cloud native design principles you use; probably never a yes/no answer
- running on a cloud and taking advantage of that (LCM, multiple workloads, etc.; possible with VMs, almost necessary with k8s)
- running on a cloud without taking advantage of that (e.g. single-VM 'add this to your VPC' solutions)
- running in a VM in many different ways

We have many terms we could use for various combinations of this and we appear to be stretching their definitions from what is commonly accepted.  How many terms do we need and how many can we unambiguously define?  (We could choose to leave out some definitions until the use cases need them.)

A better way of using the term 'cloud native network function' might be to say that one that actually uses our best practices is a cloud native network function.  Which means the glossary does not define its specific meaning; the best practices do.

doc/glossary.md
sishbi: @iawells
Isn't a monolithic network application running on a physical host a 'PNF' ?
I don't think it is a 'VNF' as it isn't virtualised or orchestrated.

tliron: Well, it could be a single application or several. There are simple PNFs and complex PNFs.

The reason why this gets complicated is that we're seeing "smarter" PNFs that can be configured to run various internal services. They can come with complex managemenet software. They can participate in cloud fabrics. Hence my comment about what it *may* comprise internally. These aspects *could* be advertised.

But, the key words for the definition are indeed hardware, discrete, and encapsulation.

doc/glossary.md
electrocucaracha: Maybe it's important to mention that in the definition, 
```suggestion
A CNF has different meanings but for this Working Group we have defined as:
**Cloud-Native Network Function (CNF)**: A network function that is a [cloud-native application](https://github.com/cncf/glossary/blob/main/definitions/cloud_native_apps.md), i.e. developed using cloud-native principles. 
```

doc/glossary.md
electrocucaracha: Agree, the term is not used by any document

doc/glossary.md
electrocucaracha: I'm not sure if this definition helps to clarify the difference between PNF, VNF and CNF. If I understand this definition a PNF will be considered CNF when implements cloud-native principles and it's supported by commodity hardware, isn't it?

doc/glossary.md
electrocucaracha: So a Cloudified NF can use Virtlet or Kubervirt technologies to provide that hybrid implementation?

doc/glossary.md
electrocucaracha: All the Containerized NFs are VNFs but not all VNFs are Containerized NFs?

doc/glossary.md
tliron: I think that because a PNF is encapsulated and discrete it immediately precludes it from cloud-native principles. It could never be a CNF.

But part of the problem is that we're not clear on what those cloud-native principles are...

doc/glossary.md
tliron: Correct. We're seeing quite a lot of use of things like KubeVirt and Kata containers in Kubernetes in telco. So, a KubeVirt-based NF would be a KNF (and thus a Cloudified NF) but _not_ a containerized NF. Whether our not it's a Cloud-native NF depends on how well its designed.

doc/glossary.md
tliron: @iawells I really wish we could just call them CNNFs and let CNF=Containerized. But quite a few in this group are insisting that CNFs have always meant "Cloud-native". I'm confident that this will always be confusing to outsiders.

@electrocucaracha If a Containerized NF doesn't use virtualization then I don't see how it would be a VNF. There are a few runtimes that incorporate aspects of both, notably Kata containers and GVisor. In those cases I think the NF could be considered both a Containerized NF and a VNF.

iawells: The literal meaning of 'virtual' here is 'simulated; not physical'. Virtual machines are machines that are virtual. Virtual network functions, depending on your reading, can be 'network functions that are not physical' (regardless of what they actually *are*) or 'network functions running in virtual machines' (if you read 'virtual' as shorthand for 'virtual machine'. The former reading has more logic, but the latter is usually the one people intend.

You two are each arguing one view without seeing the other, hence your disagreement.

tliron: @iawells That's not a bad interpretation -- a narrative about disaggregating software from hardware --  but it's just one. :) Many sources claim that VNF means "Virtualized Network Function". This is similar to the issue with CNF (Containerized) where some people interpret the "virtualization" to be about operational technology. If that weren't the case then CNF (Containerized) would not be a necessary term, because they would all be "VNFs" in the meaning that you are suggesting.

I think we should pick one interpretation and perhaps acknowledge that there are others.

doc/glossary.md
tliron: My thinking is that exactly because of the confusion of terms it is useful to have many so that we can precisely characterize what we are talking about in any given context.

doc/glossary.md
ildikov: I personally like this approach.

doc/glossary.md
ildikov: If I understand it correctly the group is currently setting the scope on things that run on Kubernetes, so in that sense I think it is useful. Also, I think this term currently doesn't have confusing alter-egos so we could use it to keep some discussions more clear as well.

My 2 cents.

doc/glossary.md
```diff
@@ -4,18 +4,34 @@
...
+
+**Kube-Native**: TBD
```
ildikov, line 27: I'm starting to feel that the *-native terms are getting a little counter productive. We are having so many of these popping up that they are starting to cause unnecessary specialization and fragmentation. I think we could live without this. At least for now.

tliron: I agree. I left "Kube-Native" from the previous version, but I think "Kubernetes Network Function" and "Cloud-Native" are both enough,

doc/glossary.md
```diff
@@ -4,18 +4,34 @@
 
 ### Community Definitions
 
-**Cloud Native** - TBD
+**Cloud-Native**: TBD
```
xmulligan:
```suggestion
**Cloud Native**: TBD
```

xmulligan: https://github.com/cncf/foundation/blob/master/style-guide.md

doc/glossary.md
xmulligan:
```suggestion
**Cloud Native Network Function (CNF)**: A network function that is a [cloud-native application](https://github.com/cncf/glossary/blob/main/definitions/cloud_native_apps.md), i.e. developed using cloud native principles. 
```

doc/glossary.md
xmulligan: +1

doc/glossary.md
xmulligan:
```suggestion
**Containerized network function**: A network function deployed entirely or primarily as one or more containers. Should not be referred to by the "CNF" acronym in order to avoid confusion with Cloud native Network Function (CNF). Note that a containerized network function is not necessarily a Kubernetes Network Function (KNF) as it is possible to use containers on other platforms. It is also not necessarily a cloudified network function. Also note that some containerization technologies include virtualization features (see also: VNF).
```

doc/glossary.md
xmulligan: Does it have to be "cloudified" to run on Kubernetes?

doc/glossary.md
```diff
@@ -4,18 +4,34 @@
 
 ### Community Definitions
 
-**Cloud Native** - TBD
+**Cloud Native**: TBD
 
-**Cloud Native Network Function** - A cloud native network function (CNF) is a [cloud native application](https://github.com/cncf/glossary/blob/main/definitions/cloud_native_apps.md) that provides network functionality, which is developed using cloud native principles. 
+**Cloud-Native Network Function (CNF)**: A network function that is a [cloud-native application](https://github.com/cncf/glossary/blob/main/definitions/cloud_native_apps.md), i.e. developed using cloud-native principles.
```
iawells:
```suggestion
**Cloud Native Network Function (Cloud Native CNF)**: A network function that is a [cloud-native application](https://github.com/cncf/glossary/blob/main/definitions/cloud_native_apps.md), i.e. developed using cloud-native principles. 
```

doc/glossary.md
iawells:
```suggestion
**CNF**: A 'cloud native network function'; by the CNCF's definition, a network function designed to run inside containers on Kubernetes as a [cloud native application](https://github.com/cncf/glossary/blob/main/definitions/cloud_native_apps.md).

This is sometimes interpreted to mean 'containerized network function'.  We do not use this definition.

Comment (jeffsaelens) [reply doc/glossary.md]: I don't think we call out designed to run in containers in the main definition. You could put it in the disclaimer below that is the packaging model we are examining first. Much like the VNF discussion around a VM being involved vs it being about abstraction, the concept of using principles to define what makes it a cloud native network function is important.

Comment (iawells) [reply doc/glossary.md]: We didn't, no - but I noted we're using a CNCF definition (per a conversation with @taylor, this term was coined by Don at some point) and, as is fairly standard with most things CNCF, we can accept that implies 'in containers on k8s'.

We know that, as a CNCF WG, we're going to be focussing on k8s apps until something drastic changes, and if it does then we'll have to do some editing.  But it seems better to accept the shorthand rather than having to qualify which specific kind of CNF we mean every time we use it.

(Yes, this is me changing sides on the argument.  I feel that if we spell it out this way in the glossary we're explaining ourselves quite clearly; and the idea of saying 'kube-native CNF' every time is giving me horrors.)

File: doc/glossary.md (hunk 2 above)
Comment (pgoyal01) [reply doc/glossary.md]: ETSI GS NFV 003 V1.4.1 defines, "Network Function (NF): functional block within a network infrastructure that has well-defined external interfaces and well-defined functional behaviour."

File: doc/glossary.md (hunk 4 above)
Comment (pgoyal01) [reply doc/glossary.md:18]: ETSI GS NFV 003 V1.4.1 defines a PNF in term of a Network Function, "Physical Network Function (PNF): implementation of a NF via a tightly coupled software and hardware system."

File: doc/glossary.md (hunk 1 above)
Comment (taylor) [reply doc/glossary.md]: @xmulligan added because of CNCF style guide https://github.com/cncf/foundation/blob/master/style-guide.md#1-cloud-native-and-open-source

@tliron will fix the other places where dash is used.

File: doc/glossary.md (hunk 11 above)
Comment (taylor) [reply doc/glossary.md]: CNCF style guide https://github.com/cncf/foundation/blob/master/style-guide.md#1-cloud-native-and-open-source

File: doc/glossary.md (hunk 8 above)
Comment (tliron) [reply doc/glossary.md]: @xmulligan I think so, Kubernetes is a cloud platform.

Comment (taylor) [reply doc/glossary.md]: +1 for CNF vs cloudified

Comment (tliron) [reply doc/glossary.md]: Changed back to cloudified.

File: doc/glossary.md (hunk 3 above)
Comment (tliron) [reply doc/glossary.md:16]: Remaining for future commits to decide.

File: doc/glossary.md (hunk 2 above)
Comment (tliron) [reply doc/glossary.md]: Incorporated this suggestion.

File: doc/glossary.md (hunk 1 above)
Comment (tliron) [reply doc/glossary.md]: For now we will keep the existing definition (cloud-native network function).

File: doc/glossary.md [hunk 12]
```diff
@@ -2,20 +2,37 @@
...
+
+**Containerized network function**: A network function deployed entirely or primarily as one or more containers. Should not be referred to by the "CNF" acronym in order to avoid confusion with Cloud Native Network Function (CNF). Note that a containerized network function is not necessarily a Kubernetes Network Function (KNF) as it is possible to use containers in other and simpler platforms. Thus, it is also not necessarily a cloudified network function. Also note that some containerization technologies include virtualization features (see also: VNF).
+
+**Kubernetes Network Function (KNF)**: A cloudified network function deployed on Kubernetes. Note that a KNF is not necessarily a containerized network function as there are ways to employ virtualization features in containers running in Kubernetes, and indeed to run full virtual machines.
```
taylor: The word cloudified does not seem necessary.

```suggestion
**Kubernetes Network Function (KNF)**: A network function deployed on Kubernetes. Note that a KNF is not necessarily a containerized network function as there are ways to employ virtualization features in containers running in Kubernetes, and indeed to run full virtual machines.
```

doc/glossary.md
```diff
@@ -2,20 +2,37 @@
...
 
-**Kubernetes** - Kubernetes is a portable, extensible, open-source platform for managing containerized workloads and services, that facilitates both declarative configuration and automation. It has a large, rapidly growing ecosystem. Kubernetes services, support, and tools are widely available. [Kubernetes.io](https://kubernetes.io/)
 
 ### CNF WG Specific Definitions
 
-**Kube-Native** - TBD
+**Cloudified network function**: A network function that designed to be deployed on cloud platforms, including Kubernetes (see: Kubernetes Network Function). A cloudified network function can use virtualization features (in which case it would also be considered to be a VNF) as well as containers (in which case it is also be considered to be a containerized network function). A cloudified network function is not necessarily a Cloud Native Network Function (CNF). To be classified as a CNF it must adhere to cloud native design principles.
```
taylor: Cloudified does not have a definition and sounds like a product buzz word.

I do not think this definition is needed.  Suggestion: remove completely
```suggestion

```

doc/glossary.md
```diff
@@ -2,20 +2,37 @@
...
+
+**Kubernetes**: [Kubernetes.io](https://kubernetes.io/) is a portable, extensible, open-source platform for managing workloads and services, that facilitates both declarative configuration and automation. It has a large, rapidly growing ecosystem. Kubernetes services, support, and tools are widely available.
+
+**Network function**: A functional block within a network infrastructure. Network functions may be components of network services.
+
+**Network service**: TBD
 
-**Cloud Native Network Function** - A cloud native network function (CNF) is a [cloud native application](https://github.com/cncf/glossary/blob/main/definitions/cloud_native_apps.md) that provides network functionality, which is developed using cloud native principles. 
+**Physical Network Function (PNF)**: A network function distinguished by its encapsulation and isolation in discrete hardware. Internally it may use various technologies, including virtualization, containerization, cloud platforms, etc.
```
taylor, line 18: +1 for keeping PNF definition as originally suggested by @tliron:
> **Physical Network Function (PNF)**: A network function distinguished by its encapsulation and isolation in discrete hardware. Internally it may use various technologies, including virtualization, containerization, cloud platforms, etc.

doc/glossary.md
```diff
@@ -2,20 +2,37 @@
...
 ### Community Definitions
 
-**Cloud Native** - TBD
+**Cloud Native**: TBD
+
+**Cloud Native Network Function (CNF)**: A network function that is a [cloud native application](https://github.com/cncf/glossary/blob/main/definitions/cloud_native_apps.md), i.e. developed using cloud native principles. 
+
+**Kubernetes**: [Kubernetes.io](https://kubernetes.io/) is a portable, extensible, open-source platform for managing workloads and services, that facilitates both declarative configuration and automation. It has a large, rapidly growing ecosystem. Kubernetes services, support, and tools are widely available.
+
+**Network function**: A functional block within a network infrastructure. Network functions may be components of network services.
```
taylor: This definition should be expanded and not dependent on definitions from other sources.  

Redacting this suggestion as of July 19th, 2021. (See the new suggestion below https://github.com/cncf/cnf-wg/pull/171/files#r662639819)

_**Network function**: A unit of network functionality which may be coupled or decoupled from underlying hardware. Network functions may provide network services directly or be components of larger network services._


While, I have tried to expand and address, there are problematic items to note including:
- network infrastructure definition is mentioned not defined
- network functions or specifically services providing the network functionality of network functions may not be running/implemented within "network infrastructure" (at least from some POV).
- intent of definitions to help with other definitions and the goals of the working group

NSM and some other groups have approached terms by getting to the root purpose and not using the existing terms which have baggage, including ~~marketing and sales purposes~~ context and purposes that are not clear to the reader. I believe we are running into that.  

We need to keep the current goal for this working group in mind:
- Cloud native and Kubernetes native best practices for running networking software with Kubernetes

The purpose of a network function at its core (dropping politics and such) is to provide some type of networking functionality -- packet inspection and filtering on the edge of network, bgp speaker, NAT 64 routing, etc. 

The network function is tech like any other tech, which can be virtualized/containerized, implemented to follow cloud native.

tliron: This is not about neither politics nor marketing. The "baggage" is common understandings in the industry, including standards bodies like ETSI, to which many of our companies belong. I don't think anybody is confused about the goal of this group, but we and others are definitely confused about terminology.

This NF definition has already gone through a few iterations in the group and I thought we reached a consensus, more or less. We can discuss again next meeting.

doc/glossary.md
tliron: I want to explain why I think a term like this is needed.

The issue is that we often want to differentiate between the technological aspects and the platform aspects. Quite a few of us are working on containerized solutions that do not involve any cloud platform. There are places where even setting up a Kubernetes cluster adds too much complexity, even for super-low-profile clusters such as k3s. Hence the need to refer to network functions that have been designed to be deployed on clouds as different from ContainerizedNFs and VNFs. This is the first step towards "cloud native": a network function could be cloudified but not yet cloud native.

(It indeed bleeds into a different interpretation of VNF that Ian mentioned -- that "virtual" refers not to the technology but to the fact that the NF software has been decoupled from hardware and can now run in various environment.)

As for "cloudification" sounding like a buzz word to you -- well, it's official terminology used by O-RAN, of which many of us are members.

tomkivlin: I can't remember if the definition was ever used, but there was for some time in this group the idea of "cloud native" being a sliding scale. Maybe it was part of the conformance work, but I thought we had wording to deal with "fully cloud native" and "not very cloud native"?

tliron: Yes, I recall that discussion. And indeed this PR is my attempt to assist. That is the entire spirit of this PR.

I personally am uncomfortable with sliding scale definitions -- can we actually enumerate all the criteria for cloud native and evaluate a network function according to them to produce a weighted score? -- and thus I propose a set of concrete definitions that are binary. Either a NF is cloudified or it is not, there is no middle ground in terms of the definition. (At least none that I can see.)

doc/glossary.md
tomkivlin: I don't quite understand the part about baggage/marketing/sales but I agree the proposed change is much clearer.

doc/glossary.md
tliron: Resolving this discussion. See the discussion about "Cloudified Network Function", which will define the fate of this term.

doc/glossary.md
tliron: The PR edit is the result of in-meeting discussions. Since this is foundational I suggest returning it to the in the in-person meeting again for discussion.

I'm OK with either definition. My one quibble is the phrase "in hardware or software". That seems like an intended reference to PNFs. However, it seems inaccurate if not misleading to me. I know of no relevant network function that can be implemented purely in hardware, unless you consider a token ring hub from 1985 to be a "network function". Even Ethernet hubs (not switches!) had signal processing that required firmware. In our more realistic context, there is always and forever a software component. All network functions, for a very long time, have required software and hardware. I would go so far as to say that the use of the word "function" already implies a soft component.

I think that if we remove "implemented in hardware or software" no meaning is lost and the definition will be "shorter and sweeter" as we always prefer.

pgoyal01: ETSI GS NFV 003 v1.4.1 (2018-08) defines:
Network Function (NF): functional block within a network infrastructure that has well-defined external interfaces and well-defined functional behaviour.

jeffsaelens: tweaked the wording slightly to make it more clear that we aren't talking about purpose built logical circuits handling all of the network functionality sans software.

taylor: @pgoyal01 is there a definition of "functional block?"

taylor: @jeffsaelens I'm not sure where your tweaks are :/

taylor: @tliron I adjusted my suggestion some based on the hardware stuff you mention.  I also redact my baggage comments (crossed out).

I find the NF definition problematic for this group. FYI, I do not find it problematic in the context of NFV spec and other places such as a 3GPP specification.  I find it conflicting and problematic specifically with talking about CNFs and Cloud native (or maybe I should even say Kubernetes native) applications and platforms. 

This group's primary goal is to identify best practices for creating and consuming networking applications running on Kubernetes. Ideally the benefits seen in enterprise app and platform development which uses the K8s stack and practices will be obtained for CNF creators and consumers.  These networking applications which follow best practices are what we call Cloud Native Network Functions (CNFs). 

When referring to a networking application or a workload with a networking application running on K8s, its good to keep in mind the definition of a [Kubernetes workload](https://kubernetes.io/docs/concepts/workloads/) which is in application, whether 1 or multiple components, running on K8s. This could be a single container in a single Pod or many of both running across multiple Nodes.

Taken this groups goal and the above information let's look at NF the definition:

> **Network function**: A functional block within a network infrastructure. Network functions may be components of network services.

- **functional block** - what is meant by functional block?  I tried to address that in my suggestion. As you say function implies software, but its treated as if this is a special thing reserved for a special place (eg. network infrastructure).  it is just software providing some requested functionality. 
- **network infrastructure** - a CNF may provide its requested functionality to anyone who can access its service end point.  why state that an NF is within network infrastructure? What if it runs at a layer with other software? This seems to be baking the historical implementation and deployment into the definition. 
- **network services** - what is a network service?  If we look at the goal of providing easy to consume services which work the same for their lifecycle as any other K8s application then a CNF could be a _simple_ application or a _complex_ application and both will provider a network service.   [linkerd layer 4 and layer 3 load balancing](https://linkerd.io/2.10/reference/architecture/) provides a network service.  An application working at lower layers of the OSI stack should ideally not be treated or named differently.


We should be able to build from the Network Function definition to a Cloud Native Network Function definition in our cloud native glossary.  Let's see how that looks

**Cloud Native Network function**: A functional block within a network infrastructure which is implemented with cloud native principles and practices. Network functions may be components of network services.

Maybe that looks okay if you dropped this into a glossary and context with all the other things known. (eg. 3GPP specifications for a mobile core component). 

Within in the broader CNCF and K8s community that definition is lacking and has all the problems I outlined above.  We are using specialized terms (noted above) instead of communicating the basic intent. Also, as I said previously, the long term intent is to -- the extent possible -- treat these networking applications like any other applications running in a K8s environment (so that all the benefits it offers can be gained).

I do not deny that this NF definition is not well known and the standard used in the networking/telco community. I am saying it is not helpful here and causes confusion (with regards to CNF and our goals).

I suggest that if we want to have a ETSI definition then lets label as such similar to what @pgoyal01 [does above](https://github.com/cncf/cnf-wg/pull/171/files#r660604900)

doc/glossary.md
```diff
@@ -2,20 +2,37 @@
 
 ## Specification
 
+
 ### Community Definitions
 
-**Cloud Native** - TBD
+**Cloud Native**: TBD
+
+**Cloud Native Network Function (CNF)**: A network function that is a [cloud native application](https://github.com/cncf/glossary/blob/main/definitions/cloud_native_apps.md), i.e. developed using cloud native principles.
```
taylor, line 10: @tliron my problems about NF are related to this definition for CNF.  a CNF is a cloud native application.  A cloud native application is an application taking advantage of cloud native innovation and so forth.  (see new cncf glossary definition and the [principles](https://networking.cloud-native-principles.org/cloud-native-networking-preamble).

doc/glossary.md
taylor: Relating this to the currently suggested CNF definition which says the CNF is a cloud native **application**.

```suggestion
**Network function**: An application providing a unit of network functionality. The network function  may be coupled or decoupled from underlying hardware. The network functionality may be provided directly as a network service or be components of larger network services.
```

An example of an application providing IPv6-only clients to contact IPv4 servers (NAT64). This could be provided as a network service directly.

The NAT64 application could be combined with another application to provided a more "complex" / larger service.

tliron: I tend to agree, except ... the term itself is alien to us in the first place. "Functional block" might not seem clear, but then is "function" any clearer? Why are network functions called functions and not applications or service? Oh, right, "network  service" is something else. And how is a function different from a network service again?

I think that there are certain terms that we cannot own and they will always exist in some degree of tension with cloud native principles. We might need to refer to accepted definitions, but perhaps what we can do is frame them within their external context and make that tension explicit (as I did in the definition of "containerized network function").

Perhaps in the future, as the industry as a whole matures, we will have a new word for these networking workloads that properly explicates their structure and role.

doc/glossary.md
iawells: Agreed (as with CNF, which requires the acknowledgement).

doc/glossary.md
iawells: We are the cloud-native network function working group by definition.  Does the best practices of non-cloudy network functions matter to us so much that we'll be using terms like this regularly?

doc/glossary.md
taylor: @tliron, We are not defining function, we are defining the term "network function".  There are definitions for the words functionality and function that do apply to the intent of the term "network function" specifically in the context of an application running K8s (aka a CNF).

Definitions:
- Function: _the action for which a person or thing is specially fitted or used or for which a thing exists_  (ref: [merriam-webster](https://www.merriam-webster.com/dictionary/function))
  - we are not referring to the math term _function_ nor a computer subroutine
- Functionality:  _the quality or state of being functional_ -- eg. a design that is admired both for its beauty and for its functionality. _especially : the set of functions or capabilities associated with computer software or hardware or an electronic device_ -- eg. data management functionalities such as data integrity, security, recoverability and manageability).  (ref: [merriam-webster](https://www.merriam-webster.com/dictionary/functionality))

The definition I [suggested above](https://github.com/cncf/cnf-wg/pull/171/files#r662639819) for Network Function, builds on the above dictionary definitions of those words
> An application providing a unit of network functionality. The network function  may be coupled or decoupled from underlying hardware. The network functionality may be provided directly as a network service or be components of larger network services.

doc/glossary.md
taylor: The word containerized can be dropped and still communicate the essential parts of a KNF:

```suggestion
**Kubernetes Network Function (KNF)**: A network function deployed on Kubernetes. Note that a KNF is not necessarily a containerized network function as there are ways to employ virtualization features in containers running in Kubernetes, and indeed to run full virtual machines.
```

taylor: I suggest we drop this definition completely until there is driving need for this acronym. For now, in use cases, best practices, and other documentation we can explicitly say things like
- NF deployed on K8s
- NF running on K8s
if needed.

```suggestion
```

doc/glossary.md
taylor: I think we can drop cloudified from the kubernetes network function definition below https://github.com/cncf/cnf-wg/pull/171/files#r681255763

tliron: Removing definition.

doc/glossary.md
tliron: Removed definition.

xmulligan (approved):

iawells (approved):

iawells:
Consensus achieved on the morning meeting.

electrocucaracha (approved):

Issue #67, opened 2021-02-11
"Do not run containers with the privilege flag" as a best practice


iawells:
I believe we could add a best practice for 'don't use privilege'.  We could add a user story to justify it (something about making the platform robust to application behaviour) but I believe the best practice can stand alone.

The higher level principle is:
- Principle of least privilege (discussion https://github.com/cncf/cnf-wg/discussions/28)

See discussion https://github.com/cncf/cnf-wg/discussions/25 for not running containers with privilege flag.

electrocucaracha:
@iawells correct  me if I'm wrong, but this user story will only cover the [Pod definition](https://kubernetes.io/docs/concepts/workloads/pods/#privileged-mode-for-containers)? are you considering to include the [execution of services with a non-root user practice ](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#user). Maybe it would be worth it to mention some efforts like [usernetes](https://github.com/rootless-containers/usernetes), [distroless](https://github.com/GoogleContainerTools/distroless) and [katacontainers](https://katacontainers.io/) which can be related to that topic.

taylor:
Related Github discussion https://github.com/cncf/cnf-wg/discussions/28

iawells:
I was thinking specifically about the k8s-granted privilege use case here.  What you're talking about are probably not going to turn up in the same best practice, nor be driven by the same use case.

I'm trying to stick to the rules that that the design is dictated by the requirements, but the requirements should not dictate unnecessary design choices.

- The user story says what you're trying to achieve and why (in this case I think we would go with 'a team is responsible for its own mistakes and an operator must be able to find the responsible team' - so an app breaking the platform through using privilege confuses that, because the app team breaks the platform and the platform team gets blamed)

- A design would suggest a way of implementing something that solves the user story, and I think a number of your suggestions might work there (along with other options).  It might or might not cover only the pod definition, and it might consider things like katacontainers.  It will almost certainly speak of pod privileges (both system-root and CAP_*), since pods with that sort of privilege can always break platform components, and that should not be possible.

  I think the 'non-root user' part is not going to work in response to the above user story, though.  It's more a 'least privilege' security requirement directed at the CNF developer in order to make more secure CNFs - so it would be driven by a different user story than the example I gave.  It doesn't actually grant you any more power over the platform or make you any more destructive if you have the in-container UID 0 (which typically is not the system root).

[...]

PR #182, opened 2021-07-22
Best practice: Container should execute process(es) as a non-root user

taylor:
This covers the least privilege best practice of avoiding running processes as root in containers

- Issue: https://github.com/cncf/cnf-wg/issues/186

See least privilege issue #67 and discussion https://github.com/cncf/cnf-wg/discussions/20

Alternative best practice name: "Container should execute process(es) as non-root user"

Co-authored-by: Ian Wells <iawells@cisco.com>


vukg:
/lgtm
one of very important practices that we for example enforce at admission

cbpps/0002-no-root-in-containers.md
```diff
@@ -0,0 +1,154 @@
...
+
+Both of these benefit the CNF developer directly, improving the quality of the CNF.  Validation of the best practice is the responsibility of the CNF developer.
+
+Indirectly, improved CNF quality benefits CNF operators.  The proposed tests are runnable by CNF operators as acceptance tests.
+
+SE-Linux based environments will require dropping root privileges. Example: OpenShift
+
+### Goals
+
+Avoiding the root user in containers improves the security and behaviour of applications when bugs occur.  It is also part of a defense in depth strategy against external compromises.
```
electrocucaracha: IMHO Application's bugs is a generic term which cover many aspects. AFAIU, the main idea is to reduce the attack surface by limiting the permissions.

```suggestion
Avoiding the usage of root user in containers improves the security reducing the attack surface on vulnerabilities .  It is also part of a defense in depth strategy against external compromises.
```

cbpps/0002-no-root-in-containers.md
```diff
@@ -0,0 +1,154 @@
...
+
+This BP recommends that the application not use the UID that can override all protections.  This means that file read and write protections can be established.
+
+It does not consider what filesystem write permissions should be in order to benefit from those protections.  CNF developers will additionally want to ensure filesystem permissions are tightened up appropriately so that non-root users are prevented from doing damage.  This is outside the BP scope.
+
+## Proposal
+
+When building a container, the container should be build to run its processes as a non-root user.  setsid processes should not be required to do the work inside a container.
+
+The root user in a container has fewer rights and is distinct from the platform's root.  The containerization process included using a user cgroup, which means that the containers' users are distinct from the platform's users and do not include most dangerous elevated rights.
```
electrocucaracha: > containers' users are distinct from the platform's users

This is true as long as the Container Runtime is using user namespaces to map the users

https://docs.docker.com/engine/security/userns-remap/

cbpps/0002-no-root-in-containers.md
```diff
@@ -0,0 +1,154 @@
...
+- K8s 11 Ways Not to Get Hacked: [8. Run Containers as a Non-Root User](https://kubernetes.io/blog/2018/07/18/11-ways-not-to-get-hacked/#8-run-containers-as-a-non-root-user)
+- Red Hat PDF [A PRACTICAL INTRODUCTION TO CONTAINER SECURITY](https://www.redhat.com/files/summit/session-assets/2017/L99901-apracticalintroductiontocontainersecurity_labguide.pdf) - 2017/05
+
+
+### Alternatives
+
+These are not strictly alternatives as they can be used with non-root, but can be applied to a container running as root.
+
+- Disable all capabilities or limit them
+- Do not run the container is privileged
```
electrocucaracha: What about using [rootless](https://www.docker.com/blog/experimenting-with-rootless-docker/) solutions like [usernetes](https://github.com/rootless-containers/usernetes)? Or using another runtime like [Kata Containers](https://katacontainers.io/)?

electrocucaracha (commented):
Are "no root" and "non-root" terms the same? If that's true the document uses those interchangeably so I'll be nice to clarify it.

cbpps/0002-no-root-in-containers.md
electrocucaracha: Apparently this [KEP](https://github.com/kubernetes/enhancements/pull/2101) is going to address some of this issues as well

iawells: rootless installs of k8s are interesting but not what we're trying to address here, this is specifically about the use of container root in apps and that's about the use of system root in the platform.

Kata containers (and clearcontainers) are interesting side cases; they're not using 'containers' in the conventional case, a constrained use of the host kernel, but looking at lightweight virtualisation to run container workloads in individual VMs.  Testing the 'don't use root' would certainly work differently there because there's no way that the container runtime isn't going to use root somewhere - but if it's compatible with container payloads all the reasoning and recommendations should still apply.

I need to read the user namespace (KEP) thing in more depth.  I'm a bit confused by it, containers typically use a user namespare already and I don't quite understand the distinction.  On the other hand, we write with what we have, and we don't have KEP yet, so this document as-is is probably a valid recommendation for k8s's you can deploy today.

cbpps/0002-no-root-in-containers.md
```diff
@@ -0,0 +1,154 @@
...
+- [ ] (R) CBPP approvers have approved the CBPP status as `implementable`
+- [x] (R) CBPP summary, motivation and best practice details are appropriately documented
+- [ ] (R) Test plan is in place, giving consideration to CNF Test Suite input
+- [x] (R) Scoring has been determined
+- [x]   "Implementation History" section is up-to-date
+- [ ]    Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+## **Summary**
+
+Containers have a list of their own users independent of the host system, one of which is UID 0, the root user. Containers should run processes as a user other than root which makes it easier to run these images securely. Processes in a container should not run as root, nor should they assume they are running as root (UID 0).
```
jeffsaelens: The second sentence reads weird (flow is off).

cbpps/0002-no-root-in-containers.md
```diff
@@ -0,0 +1,154 @@
...
+ 
+
+## **Motivation**
+
+Both of these benefit the CNF developer directly, improving the quality of the CNF.  Validation of the best practice is the responsibility of the CNF developer.
```
jeffsaelens: Both of what?

cbpps/0002-no-root-in-containers.md
jeffsaelens: This isn't written as a goal.

cbpps/0002-no-root-in-containers.md
```diff
@@ -0,0 +1,154 @@
...
+
+### Non-Goals
+
+This BP recommends that the application not use the UID that can override all protections.  This means that file read and write protections can be established.
+
+It does not consider what filesystem write permissions should be in order to benefit from those protections.  CNF developers will additionally want to ensure filesystem permissions are tightened up appropriately so that non-root users are prevented from doing damage.  This is outside the BP scope.
+
+## Proposal
+
+When building a container, the container should be build to run its processes as a non-root user.  setsid processes should not be required to do the work inside a container.
```
jeffsaelens: My suggested change button is MIA, build should be built*

cbpps/0002-no-root-in-containers.md
```diff
@@ -0,0 +1,154 @@
...
+
+Unix access enforcement (user/group) will be respected. As an added advantage, fine-grained access enforcement, such as SELinux, will also hold.
+
+### User Story
+
+_TBD: the user stories for access control need writing_
+
+### Tradeoffs/Constraints/Caveats/Notes
+
+Container images are frequently built from upstream image versions made from OS deployments.  These will include things like setsid binaries as a part of their base configuration.  If CNF developers follow this best practice they will have to audit and clean up any upstream images to respect this rule (removing files, removing packages or changing permissions as appropriate).
```
jeffsaelens, line 90: This leads into another best practice around scanning and onboarding CNFs :-)

cbpps/0002-no-root-in-containers.md
```diff
@@ -0,0 +1,154 @@
...
+- Do not run the container is privileged
+
+
+## Workload Context
+
+All pod types should implement this best practice.
+
+## Test Plan
+
+Inside a compliant application, all containers will avoid use of root.
```
jeffsaelens: Do we want to flesh out here what tests we would like to see? If it's in the test suite already, then we're good to go. If it is not, we could bring those to the test suite team. This could be a way to provide the CNF-WG folks who want to write some code an opportunity to contribute automated tests to CNG-TS.

taylor: Re-writing to remove the word "compliant"

```suggestion
An application which follows this best practice will not have any containers with processes running as root
```

cc: @tliron @iawells

cbpps/0002-no-root-in-containers.md
iawells:
```suggestion
Containers have a list of their own users independent of the host system, one of which is UID 0, the root user. Containers should run processes as a user other than root which makes it easier to run these images securely.
```

cbpps/0002-no-root-in-containers.md
iawells:
```suggestion
This best practice benefits the CNF developer, improving the quality of the CNF and reducing the developer's likelihood of having to diagnose problems with the CNF.  Validation of the best practice is the responsibility of the CNF developer.
```

cbpps/0002-no-root-in-containers.md
iawells: It is not exclusively that.  Running processes as non-root means that software flaws should be identifiable earlier.  It does mean that compromises can't get so far, and that's a major part of its benefit, but it's not the only benefit.

iawells: What would a goal look like?

cbpps/0002-no-root-in-containers.md
iawells:
```suggestion
When building a container, the container should be built to run its processes as a non-root user.  setsid processes should not be required to do the work inside a container.
```

cbpps/0002-no-root-in-containers.md
iawells: Do we need to write that down as its own best practice?

cbpps/0002-no-root-in-containers.md
iawells: Acceptance testing stages should be in the receiving process, yep.  That's a process, though, not a best practice, so I wonder how we write that down.

cbpps/0002-no-root-in-containers.md
```diff
@@ -0,0 +1,154 @@
...
+
+- Disable all capabilities or limit them
+- Do not run the container is privileged
+
+
+## Workload Context
+
+All pod types should implement this best practice.
+
+## Test Plan
```
iawells:
```suggestion
## Test Objectives
```

iawells: (per @jeffsaelens' other pull request and comments)

cbpps/0002-no-root-in-containers.md
taylor: @electrocucaracha a root process could cause unintended corruption or delete files that accidentally, both of which might be considered general bugs vs security specific.  This is part of why we did not keep this practice specific to security.

cbpps/0002-no-root-in-containers.md
taylor: Seems like the user namespace mapping would be good additional best practice.

@iawells I think the suggestion is to change the phrasing slightly to not say users will be distinct. @electrocucaracha correct me if I am misunderstanding.

Here is an update based on that understanding:

```suggestion
The root user in a container has fewer rights and is distinct from the platform's root.  The containerization process includes using a user cgroup, which means that the containers' users can be distinct from the platform's users and do not include most dangerous elevated rights. An exception to that separation is if the container runtime is not using user name spaces to map the users.
```

cbpps/0002-no-root-in-containers.md
taylor:
```suggestion
Avoiding root in containers can help to:
- Improve the security and behaviour of applications.
- Add to the defense in depth strategy against external compromises.
```

taylor: @jeffsaelens are you looking for something worded more like that ☝🏻

cbpps/0002-no-root-in-containers.md
taylor: @jeffsaelens add to a new or existing discussion https://github.com/cncf/cnf-wg/discussions.

Maybe something like:
- Applications should be auditable for security vulernabilities when onboarding by a receiving party.

cbpps/0002-no-root-in-containers.md
jeffsaelens: Yes 👍🏻

cbpps/0002-no-root-in-containers.md
jeffsaelens:
```suggestion
When building a container, the container should be built to run its processes as a non-root user.  setsid processes should not be required to do the work inside a container.
```

And suddenly I can suggest edits.

cbpps/0002-no-root-in-containers.md
electrocucaracha:
```suggestion
The root user in a container has fewer Linux Kernel capabilities and may be distinct from the platform's root (if the container runtime enables user namespaces remap feature).
```

https://docs.docker.com/engine/security/#linux-kernel-capabilities
https://d3oypxn00j2a10.cloudfront.net/assets/img/Docker%20Security/WP_Intro_to_container_security_03.20.2015.pdf

AFAIK Cgroups allow you to allocate resources among user-defined groups of tasks (processes) running on a system.

iawells: It's confusing.  You're correct, and this wants fixing - it's a user namespace, not a cgroup, so we used the wrong word; but namespaces and (some) cgroups (for instance, the PID cgroup) do very similar things.

cbpps/0002-no-root-in-containers.md
```diff
@@ -0,0 +1,154 @@
...
+
+By default, the first process starting in a container runs as root - you have to actively take steps to shed the permissions (Dockerfile USER line, plus installing files with appropriate ownership within the Dockerfile).
+
+We specifically want to run unprivileged process so that their access is limited.  Of course, if access is limited then the CNF developer must ensure that access is still available to the things the CNF is going to need to access.  This will involve changing permissions on data files and on working directories when the container image is constructed (they cannot be changed on startup because the application does not have the right to do that).  This may also affect the use of shared volume mounts or host mounts - ownership of and rights on the root directory must permit access to the container users.
+
+### References
+
+- CNF WG discussion: https://github.com/cncf/cnf-wg/discussions/20
```
electrocucaracha, line 98: Maybe this [CVE](https://seclists.org/oss-sec/2019/q1/119) has to be referred here

wvwatson (commented):
+1

cbpps/0002-no-root-in-containers.md
```diff
@@ -0,0 +1,154 @@
...
+
+Indirectly, improved CNF quality benefits CNF operators.  The proposed tests are runnable by CNF operators as acceptance tests.
+
+SE-Linux based environments will require dropping root privileges. Example: OpenShift
+
+### Goals
+
+Avoiding the root user in containers improves the security and behaviour of applications when bugs occur.  It is also part of a defense in depth strategy against external compromises.
+
+Avoid compromised apps causing more damage.
```
taylor:
```suggestion
- Avoid compromised apps from causing more damage.
```

cbpps/0002-no-root-in-containers.md
pgoyal01: "A container's root user has fewer Linux Kernel permissions and may be distinct from the platform's root (if the container runtime enables user namespaces remap feature)."

cbpps/0002-no-root-in-containers.md
```diff
@@ -0,0 +1,155 @@
...
+
+This BP recommends that the application not use the UID that can override all protections.  This means that file read and write protections can be established.
+
+It does not consider what filesystem write permissions should be in order to benefit from those protections.  CNF developers will additionally want to ensure filesystem permissions are tightened up appropriately so that non-root users are prevented from doing damage.  This is outside the BP scope.
+
+## Proposal
+
+When building a container, the container should be built to run its processes as a non-root user.  setsid processes should not be required to do the work inside a container.
+
+The root user in a container has fewer Linux Kernel capabilities and may be distinct from the platform's root (if the container runtime enables user namespaces remap feature).
```
pgoyal01:
```suggestion
A container's root user has fewer Linux Kernel privileges and may be distinct from the platform's root (if the container runtime enables user namespaces remap feature).
```

cbpps/0002-no-root-in-containers.md
taylor: yes to adding to the references @electrocucaracha

cbpps/0002-no-root-in-containers.md
```diff
@@ -0,0 +1,155 @@
...
+- K8s Documentation - Securing a Cluster: [Controlling what privileges containers run with](https://kubernetes.io/docs/tasks/administer-cluster/securing-a-cluster/#controlling-what-privileges-containers-run-with)
+- [Security Best Practices for Kubernetes Deployment](https://kubernetes.io/blog/2016/08/security-best-practices-kubernetes-deployment/) - 2016/08/31
+- [Docker and Kubernetes — root vs. privileged](https://itnext.io/docker-and-kubernetes-root-vs-privileged-9d2a37453dec) - 2020/06/25
+- Bitnami on [Use Non-Root Containers](https://docs.bitnami.com/kubernetes/faq/configuration/use-non-root/) - 2020/05/13
+- [Running non-root containers on Openshift](https://engineering.bitnami.com/articles/running-non-root-containers-on-openshift.html)
+- K8s 11 Ways Not to Get Hacked: [8. Run Containers as a Non-Root User](https://kubernetes.io/blog/2018/07/18/11-ways-not-to-get-hacked/#8-run-containers-as-a-non-root-user)
+- Red Hat PDF [A PRACTICAL INTRODUCTION TO CONTAINER SECURITY](https://www.redhat.com/files/summit/session-assets/2017/L99901-apracticalintroductiontocontainersecurity_labguide.pdf) - 2017/05
+
```
electrocucaracha, line 113:
```suggestion
- [CVE-2019-5736: runc container breakout](https://seclists.org/oss-sec/2019/q1/119)

```

cbpps/0002-no-root-in-containers.md
```diff
@@ -0,0 +1,155 @@
...
+
+### Alternatives
+
+These are not strictly alternatives as they can be used with non-root, but can be applied to a container running as root.
+
+- Disable all capabilities or limit them
+- Do not run the container is privileged
```
pgoyal01:
```suggestion
- Do not run the container in privileged mode
```

cbpps/0002-no-root-in-containers.md
```diff
@@ -0,0 +1,155 @@
...
+- K8s Documentation - Securing a Cluster: [Controlling what privileges containers run with](https://kubernetes.io/docs/tasks/administer-cluster/securing-a-cluster/#controlling-what-privileges-containers-run-with)
+- [Security Best Practices for Kubernetes Deployment](https://kubernetes.io/blog/2016/08/security-best-practices-kubernetes-deployment/) - 2016/08/31
+- [Docker and Kubernetes — root vs. privileged](https://itnext.io/docker-and-kubernetes-root-vs-privileged-9d2a37453dec) - 2020/06/25
+- Bitnami on [Use Non-Root Containers](https://docs.bitnami.com/kubernetes/faq/configuration/use-non-root/) - 2020/05/13
+- [Running non-root containers on Openshift](https://engineering.bitnami.com/articles/running-non-root-containers-on-openshift.html)
+- K8s 11 Ways Not to Get Hacked: [8. Run Containers as a Non-Root User](https://kubernetes.io/blog/2018/07/18/11-ways-not-to-get-hacked/#8-run-containers-as-a-non-root-user)
+- Red Hat PDF [A PRACTICAL INTRODUCTION TO CONTAINER SECURITY](https://www.redhat.com/files/summit/session-assets/2017/L99901-apracticalintroductiontocontainersecurity_labguide.pdf) - 2017/05
+
+
+### Alternatives
```
taylor:
```suggestion
### Alternatives / Related Practices
```

tliron: I think this section's purpose is unclear. We *do* need a lot more discussion on:
* You think you need to root in order to achieve X, but actually you don't! There's a specific solution Y that can achieve what you need without root.
* When you do need root (there is no solution Y), what can you do to minimize the risks?

cbpps/0002-no-root-in-containers.md
taylor: The CNF Test Suite is looking to use [Falco's user_must_run_as_non_root test](https://github.com/falcosecurity/falco/blob/master/test/psps/user_must_run_as_non_root.yaml). The ticket for adding the test is https://github.com/cncf/cnf-testsuite/issues/920

cbpps/0002-no-root-in-containers.md
taylor: Added the info you suggested @electrocucaracha

taylor:
Alternative best practice name: "Container should execute process as non-root user"

cbpps/0002-no-root-in-containers.md
iawells: We've listed here what we'd like to test for, and we've listed a Falco test identifier with no indication of what it's actually testing for.  Can we make the two meet in the middle?

(We will also need to specify a given version of the Falco test, since future versions might test for something different.)

iawells:
> Alternative best practice name: "Container should execute process as non-root user"

We're asking for no process ever to run as root in a container, not just the first process so we might want to phase that slightly less ambiguously (or it might be read as 'the *first* process will not be launched as root'); but 'no root in containers' is really no better.

pgoyal01:
@iawells would an alternative such as, "Container should execute a process as non-root user" or "Containers should execute processes as non-root user" or "Processes within containers should be executed as a non-root user" or ...

cbpps/0002-no-root-in-containers.md
```diff
@@ -0,0 +1,162 @@
+# CBPP-000N: No root in containers
```
taylor:
```suggestion
# CBPP-000N: Container should execute process(es) as non-root user
```

cbpps/0002-no-root-in-containers.md
```diff
@@ -0,0 +1,165 @@
...
+
+
+
+Obviously, a well-written CNF would not be attempting to do things it should not do.  But all software has bugs.  Also, executing processes can be compromised by outside forces, and if this happens filesystem protection is a part of a "Defense in depth" strategy to ensure the compromise does not escalate.
+
+Unix access enforcement (user/group) will be respected. As an added advantage, fine-grained access enforcement, such as SELinux, will also hold.
+
+### User Story
+
+_TBD: the user stories for access control need writing_
```
jeffsaelens: @jeffsaelens add to user story

denverwilliams (approved):

xmulligan (approved):

agentpoyo (approved):
LGTM

cbpps/0002-no-root-in-containers.md
```diff
@@ -0,0 +1,180 @@
...
+
+A container's root user has fewer Linux Kernel privileges and may be distinct from the platform's root (if the container runtime enables user namespaces remap feature).
+
+However, the container's root user does have full read/write access to the container's filesystem.  It can read or modify any file.  No secrets can be kept from it; it cannot be prevented from changing the content of all executable files on the system.
+
+On a basic level, avoiding the root user means that the container filesystem permissions are enforceable against all processes running in the system.  Those processes can be prevented from doing critical things like:
+
+- viewing secrets they should not be viewing
+- modifying binaries within the filesystem that will later be executed
+- What about using rootless solutions like usernetes? Or using another runtime like Kata Containers?
```
michaelspedersen: Is this supposed to be included (seems more like a WIP note)

cbpps/0002-no-root-in-containers.md
```diff
@@ -0,0 +1,180 @@
...
+Containers have a list of their own users independent of the host system, one of which is UID 0, the root user. Containers should run processes as a user other than root which makes it easier to run these images securely.
+ 
+
+## **Motivation**
+
+This best practice benefits the CNF developer, improving the quality of the CNF and reducing the developer's likelihood of having to diagnose problems with the CNF.  Validation of the best practice is the responsibility of the CNF developer.
+
+Indirectly, improved CNF quality benefits CNF operators.  The proposed tests are runnable by CNF operators as acceptance tests.
+
+SE-Linux based environments will require dropping root privileges. Example: OpenShift
```
michaelspedersen, line 40: Is this related to executing as root user or setting 'privileged: true' flag in Docker/K8s when creating the container? If it's related to the privileged flag then I would not consider this motivation by itself.

cbpps/0002-no-root-in-containers.md
```diff
@@ -0,0 +1,180 @@
...
+
+This BP recommends that the application not use the UID that can override all protections.  This means that file read and write protections can be established.
+
+It does not consider what filesystem write permissions should be in order to benefit from those protections.  CNF developers will additionally want to ensure filesystem permissions are tightened up appropriately so that non-root users are prevented from doing damage.  This is outside the BP scope.
+
+## Proposal
+
+When building a container, the container should be built to run its processes as a non-root user.  setsid processes should not be required to do the work inside a container.
+
+A container's root user has fewer Linux Kernel privileges and may be distinct from the platform's root (if the container runtime enables user namespaces remap feature).
```
michaelspedersen: would it make sense to change "privileges" to "capabilities" here (as you can add/remove individual privileges to a container, but privileged is more of a "catch-all")

cbpps/0002-no-root-in-containers.md
```diff
@@ -0,0 +1,180 @@
...
+Unfortunately, the latest resources it downloaded have been compromised on the central system. The  OpenVSwitch runs compromised code which takes over the container’s process with its own malicious code.
+
+With this best practice, the now malicious application will not be able to escalate its damage beyond what the non-root user has access to.
+
+
+#### A Firewall as a Service (FWaaS) CNF  downloads compromised externally defined rules
+
+A firewall running on a K8s cluster applies rules from an external, centralized system. These firewall rules are applied “hot” to the running container and are expected to be pre-validated by the external system.
+
+Unfortunately, the latest set of rules include an exploit while still passing validation by the external system and the firewall CNF itself. The firewall downloads the newly defined rules with the exploit and when they are applied the firewall container is compromised with malicious code.
```
michaelspedersen: I think this example is only partially relevant when talking about non-root. The updated rules could still compromise the firewall and it's functionality regardless of root user. As for compromising the container, this user story is practically identical to the one above.

cbpps/0002-no-root-in-containers.md
```diff
@@ -0,0 +1,180 @@
...
+
+By default, the first process starting in a container runs as root - you have to actively take steps to shed the permissions (Dockerfile USER line, plus installing files with appropriate ownership within the Dockerfile).
+
+We specifically want to run unprivileged process so that their access is limited.  Of course, if access is limited then the CNF developer must ensure that access is still available to the things the CNF is going to need to access.  This will involve changing permissions on data files and on working directories when the container image is constructed (they cannot be changed on startup because the application does not have the right to do that).  This may also affect the use of shared volume mounts or host mounts - ownership of and rights on the root directory must permit access to the container users.
```
michaelspedersen: A little unclear on the meaning of "unprivileged". Is this from a container or root user standpoint? Also consider changing "**their** access" in the first sentence to CNF developers

cbpps/0002-no-root-in-containers.md
ildikov: If we keep these as questions I think they should be a top level item as they don't seem to belong to this list.

michaelspedersen (commented):
Generally looks good, but I have a couple comments and questions as I am a little unclear about the use of "privileged" and "privileges" throughout the BP

taylor:
@pgoyal01 please review again. Your items have been resolved and user stories have been added.

ildikov (commented):
Hi,

I have only one nit pick in line, please check.

This patch also seem to contain quite some unrelated changes. I don't know if that's because of a rebase or it's intentionally addressing multiple issues, it would be great to avoid this in the future if possible as it makes reviewing harder.

Thanks,
Ildikó

cbpps/0002-no-root-in-containers.md
```diff
@@ -0,0 +1,155 @@
...
+- [ ] (R) CBPP approvers have approved the CBPP status as `implementable`
+- [x] (R) CBPP summary, motivation and best practice details are appropriately documented
+- [ ] (R) Test plan is in place, giving consideration to CNF Test Suite input
+- [x] (R) Scoring has been determined
+- [x]   "Implementation History" section is up-to-date
+- [ ]    Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+## **Summary**
+
+Containers have a list of their own users independent of the host system, one of which is UID 0, the root user. Containers should run processes as a user other than root which makes it easier to run these images securely.
```
pgoyal01: "these images" -- which?
```suggestion
Containers have a list of their users independent of the host system, one of which is UID 0, the root user. Containers should run processes as a user other than root which makes it easier to run these processes securely.
```

cbpps/0002-no-root-in-containers.md
pgoyal01: I think we are using "privileges"
```suggestion
The root user in a container has fewer privileges than the privileges of the platform's root (if the container runtime enables user namespaces remap feature).
```

wvwatson (approved):

cbpps/0002-no-root-in-containers.md
```diff
@@ -0,0 +1,155 @@
...
+However, the container's root user does have full read/write access to the container's filesystem.  It can read or modify any file.  No secrets can be kept from it; it cannot be prevented from changing the content of all executable files on the system.
+
+On a basic level, avoiding the root user means that the container filesystem permissions are enforceable against all processes running in the system.  Those processes can be prevented from doing critical things like:
+
+- viewing secrets they should not be viewing
+- modifying binaries within the filesystem that will later be executed
+
+Obviously, a well-written CNF would not be attempting to do things it should not do.  But all software has bugs.  Also, executing processes can be compromised by outside forces, and if this happens filesystem protection is a part of a "Defense in depth" strategy to ensure the compromise does not escalate.
+
+Unix access enforcement (user/group) will be respected. As an added advantage, fine-grained access enforcement, such as SELinux, will also hold.
```
pgoyal01: we are referring to Linux throughout
```suggestion
User/group access enforcement will be respected. As an added advantage, fine-grained access enforcement, such as in SELinux, will also hold.
```

cbpps/0002-no-root-in-containers.md
```diff
@@ -0,0 +1,180 @@
...
+
+### User Story
+
+#### A containerized OpenVSwitch downloads compromised routing updates
+
+A OpenVSwitch VNF was containerized and is running on a K8s cluster. It retrieves its routing scripts and logic are defined an externally from remote service. It then applies them to the running container instance.
```
pgoyal01: Sentence issue
```suggestion
An OpenVSwitch VNF was containerized and is running on a K8s cluster. It retrieves its routing scripts and logic, which are defined externally, from a remote service. It then applies them to the running container instance.
```

cbpps/0002-no-root-in-containers.md
```diff
@@ -0,0 +1,180 @@
...
+
+Unfortunately, the latest resources it downloaded have been compromised on the central system. The  OpenVSwitch runs compromised code which takes over the container’s process with its own malicious code.
```
pgoyal01:
```suggestion
Unfortunately, the latest resources it downloaded have been compromised. The  OpenVSwitch runs the compromised code which takes over the container’s process with its own malicious code.
```

cbpps/0002-no-root-in-containers.md
iawells: That one is awkward - poor wording choice, perhaps, but the basic thing that a root user can do with the capabilities typically granted to it (i.e. without any addition privileges added) is change the filesystem regardless of ownership and run things as any user.  There's no reason why that should be needed to run software, and it means that no file in the container is truly immutable.

cbpps/0002-no-root-in-containers.md
iawells: Agreed.  This is a flaw in the software design.  In fact, it starts saying 'rules' can be downloaded and ends up saying 'code can be run' - 'code' is not 'rules'. I don't think it's conveying what it intended to say.

It's better phrased as an argument that a container should hold all the code it's going to run and not e.g. download and run executables either (a) at all (ideally) or (b) with excessive privilege (if it must).  Firewalls are not native code, and even if they're a full programming language (typically they're not, so the example is a bit off) then the answer is sandboxing.

cbpps/0002-no-root-in-containers.md
iawells: I think we want to clear up the ambiguity by using 'non-root'.

cbpps/0002-no-root-in-containers.md
taylor: @pgoyal01 it's referring to the container images.  Maybe we should say container images?

taylor: @iawells thoughts?

cbpps/0002-no-root-in-containers.md
taylor: @michaelspedersen this best practice is only about executing processes as a non-root user.

cbpps/0002-no-root-in-containers.md
taylor: @michaelspedersen when this is applied to just the case of users running a process vs privliege=true/false for a container do you think the word choice of "privileges" works or do you think "capabilities" is still a better choice?

cbpps/0002-no-root-in-containers.md
taylor: that was accidental.  I'll remove that. thanks @michaelspedersen

taylor: These questions were a mistake and have been removed @ildikov

cbpps/0002-no-root-in-containers.md
taylor: @iawells @michaelspedersen is the suggestion to update this user story or remove it entirely and only keep the OpenVSwitch one?

iawells: You might want to keep it handy for future use but it doesn't seem to be relevant to this.  I think the more relevant use case is 'in the event that someone gets malicious code to run in this CNF you don't want the code to have elevated system rights'.  Your use case doesn't quite reflect that.

iawells: And I haven't seen the OVS one in context, but I think you'd have the same problems with it: you're trying to say 'we inject malicious code' which could make use of the system rights levels, but you're actually saying 'we inject malicious _intent_' which is unlikely to make use of those additional rights.

cbpps/0002-no-root-in-containers.md
taylor: updated @michaelspedersen @iawells

cbpps/0002-no-root-in-containers.md
pgoyal01: @taylor "container images" would be OK but since that would be first use, we should change "these" to "the"

taylor: Changed to "the container images" @pgoyal01

pgoyal01 (approved):

cbpps/0002-no-root-in-containers.md
taylor: @iawells so need to expand one of the example user stories to have something that would not be accessible without having the exploited process running as the root user.  Thoughts?

iawells: So we have two possibilities here.

This user story speaks to 'I can tell OVS to drop a firewall if I can get into its control stream and that would be bad'.  That could better be phrased as 'between the trusted input to an orchestrator or a CNF and the actual behaviour it implements is a bunch of manipulation and transport of that information, if you can get into that and change its behaviour the CNF will misbehave, possible severely'

Another user story would go something like 'I write this clever CNF where you give it a bunch of rules, it compiles those rules into machine code and ships them out to its forwarding elements.  Someone gets in there and replaces that code with code of their own.  When that code is run, it could...?' - as a non-root you still have management network access, you can still contend on poorly defended shared resources (which aren't supposed to exist, but hey, containers); but for more fun you would have to escalate privileges much more than if you were container root with the rights that go along with that.

cbpps/0002-no-root-in-containers.md
taylor: I replaced privileges with capabilities @michaelspedersen

cbpps/0002-no-root-in-containers.md
taylor: Replaced these user stories with new ones seen in https://github.com/cncf/cnf-wg/pull/193

taylor:
@ildikov  
> I have only one nit pick in line, please check.

fixed thanks.

> This patch also seem to contain quite some unrelated changes. I don't know if that's because of a rebase or it's intentionally addressing multiple issues, it would be great to avoid this in the future if possible as it makes reviewing harder.

Yes, that was a rebase. I hoped to avoid updates that were already accepted. Unfortunately, we ended up with more changes showing :/

cbpps/0002-no-root-in-containers.md
taylor: @pgoyal01 there was a request to change some of these things to capabilities so it did not cause confusion with privilege=true/false for containers.

ildikov (approved):
Thanks for the quick fix, it looks good to me now.

I also think it might be worthwhile to merge this and have follow up PRs in case there are still other bits to fix to avoid it picking up more unrelated changes.

cbpps/0002-no-root-in-containers.md
michaelspedersen: I can see you already had a good discussion. As mentioned by @iawells, the examples are likely still relevant for other BPs. I agree with replacing them with the supply chain user stories.

michaelspedersen (approved):

user-stories/supply-chain-attacks.md
```diff
@@ -0,0 +1,29 @@
...
+
+'Defence in depth' says that we should (a) defend against supply chain attacks but also (b) add mitigations in the case that supply chain attacks happen - that is, we should not assume that a single line of defence will hold.
+
+## A CNF downloads compromised updates
+
+A CNF running on a K8s cluster is downloaded from a centralized registry.  Updates also come from this registry.  The operator installs new images in the centralised repository as they receive them from the CNF developer.
+
+In this case, if the operator ensures that processes in the container will be run as a normal conatiner user without extra rights, the malicious code will have a very limited ability to do dangerous things.  It will be harder for it to try to escalate its privileges, it has no rights outside of the conatiner, it cannot change files within the container that could cause more damage (such as root-owned settings files in /etc).
```
electrocucaracha:
```suggestion
In this case, if the operator ensures that processes in the container will be run as a normal container user without extra rights, the malicious code will have a very limited ability to do dangerous things.  It will be harder for it to try to escalate its privileges, it has no rights outside of the container, it cannot change files within the container that could cause more damage (such as root-owned settings files in /etc).
```

user-stories/supply-chain-attacks.md
```diff
@@ -0,0 +1,29 @@
...
+
+Thus: limiting the privilege of container processes makes the system safer to use.
+
+## A CNF succumbs to code injection
+
+As above, a CNF runs malicious code, in this case introduced into the process through a code injection attack on its internal or external interfaces by a bad actor within the operator's organisation or by anyone on the network, respectively.  In this instance, limiting the power wielded by processes in the container again prevents the attack from going further.
+
+## A CNF succumbs to malicious instructions
+
+As an example: the configuration of a CNF might allow the operator to determine paths where a log file is written.  If this path is set to a protected file, then the conatiner root user has the rights necessary to overwrite that file.  This could be exploited to change configuration and settings files within the conatiner image and subvert it.
```
electrocucaracha:
```suggestion
As an example: the configuration of a CNF might allow the operator to determine paths where a log file is written.  If this path is set to a protected file, then the container root user has the rights necessary to overwrite that file.  This could be exploited to change configuration and settings files within the container image and subvert it.
```

electrocucaracha (changes requested):

electrocucaracha (approved):

PR #203, opened 2021-11-22
Create air gapped environments user stories

jeffsaelens:
First pass at the air gapped environment user story.


user-stories/use-caseair-gapped-environments.md
```diff
@@ -0,0 +1,25 @@
...
+
+Note: Air gapped environments are not a challenge solely for service providers/network operators. Other operators of heavily regulated environments, such as US DoD networks, seeking to employ CNFs would be beholden to these constraints.
+
+## A CNF utilizes a cloud based licensing model
+
+A CNF running on a K8s cluster is downloaded from a centralized registry. Upon instantiation, the CNF validates that it has the required license tied to a suite of desired functionality. If the license is valid, the functionality is enabled. If a valid license is not available, the desired behavior is often that the next available license in a pool is requested, or a new order is generated in real time.
+
+In this case, if the operator does not allow for a CNF to access a remote licensing server, this method breaks down. Additonally, if the licensing server is pulled into the locally hosted environments but requires licenses be purchased in advance, then the desired behavior of dynamically deployed CNFs might be hindered due to a lack of available licenses in the pool.
```
xmulligan:
```suggestion
In this case, if the operator does not allow for a CNF to access a remote licensing server, this method breaks down. Additionally, if the licensing server is pulled into the locally hosted environments but requires licenses be purchased in advance, then the desired behavior of dynamically deployed CNFs might be hindered due to a lack of available licenses in the pool. 
```

pgoyal01: How does it matter if the license server is local or remote when licenses need to be purchased in advance?

user-stories/use-caseair-gapped-environments.md
```diff
@@ -0,0 +1,25 @@
+# User stories: Employing Air Gapped Environments
+
+Environments hosting critical infrastructure often deny access to the public internet, even via a proxy. This is done for several reasons such as controlling the artifacts that come into an environment, controlling builds at the source, and ensuring malicious code is not able to call back to remote destinations. Beyond the security implications of air gapping an environment and utilizing private registries, upstream repository management becomes a centralized task that all internal stakeholders can take advantage of. Internal developers and operations teams can rely on their SREs to ensure that the images, packages, etc... they require are available and secure.
```
pgoyal01: Re: "... upstream repository management becomes a centralized task that all internal stakeholders can take advantage of" -- is it up to the stakeholders to decide or it would be mandatory?

user-stories/use-caseair-gapped-environments.md
```diff
@@ -0,0 +1,25 @@
...
+
+Note: Air gapped environments are not a challenge solely for service providers/network operators. Other operators of heavily regulated environments, such as US DoD networks, seeking to employ CNFs would be beholden to these constraints.
+
+## A CNF utilizes a cloud based licensing model
+
+A CNF running on a K8s cluster is downloaded from a centralized registry. Upon instantiation, the CNF validates that it has the required license tied to a suite of desired functionality. If the license is valid, the functionality is enabled. If a valid license is not available, the desired behavior is often that the next available license in a pool is requested, or a new order is generated in real time.
```
pgoyal01: nitpick: "... downloaded from a centralized registry" should this be "... downloaded from a centralized repository"

user-stories/use-caseair-gapped-environments.md
jeffsaelens: Maybe I didn't articulate this well enough (we might need to reword). Having to purchase in advance is equally bad. Ultimately, we need some type of true up model or the like, where I can deploy as needed, and then settle the bill at a later date. Being forced to do bulk purchases in advance, or connect to a remote licensing server both break things.

user-stories/use-caseair-gapped-environments.md
jeffsaelens: This will vary between companies and maybe even orgs. It will largely depend on the level of control desired by those in charge of security or versioning. There could be a group that just designates what packages are allowed and that's that. Or, it could be that the stakeholders just adhere to guidelines for getting their repos added and brought into the pipeline without the need to engage a third party.

user-stories/use-caseair-gapped-environments.md
jeffsaelens: Yes, it should be. Good catch.

pgoyal01:
@jeffsaelens I believe we need to consider the practicality of completely air-gapped environments -- a completely air-gapped environment wouldn't be of much use. All machines belonging to an air-gapped environment would have to be air-gapped always and that raises the issue of how to ensure that for remote machines. So, how to control the not air-gapped machines? For example, is the constraint that not air-gapped machines can not "send" or perform "put" operations on any air-gapped machines enough? In the financial sector, when connected to the corporate environment (through VPN with codes being sent to your corporate mobile phone) all access to the machine's local ports, disks, etc. is blocked. This way, you can only work using the corporate sw/tools and cannot upload any files to the corporate environment. 

What happens in hybrid cloud environments, even where the public and private clouds have a direct network interconnection as a public cloud can't be air-gapped.

iawells:
@pgoyal01:

Air-gapped here is more about the control network than machines (the machines are connected to a control network, but that control network is not connected to other networks).  It typically means that network has no connection to other networks, specifically in this case including the internet where otherwise it might go grab software from - because the best firewall is no connection at all, and the people who operate these networks want the best firewall.  It has obvious benefits - no-one can try and attack your network via network attacks if it can't be reached - and subtle ones - I know exactly where these boxes are getting software from because they have no choice.  A defence of 'it is connected to the internet, but there's a firewall' is generally not considered best practice in my experience, or at least it's less common than keeping that network completely isolated.

So:

- certain control operations (including, but not limited to, software update) come from an isolated control network
- obviously we are running CNFs, and other network connections will exist, but those connections will not accept control commands and will not be connected within the server to control software (via _name your defences here_)
- since the network is isolated, some means must be in place to get new software onto it, which should be ripe for best practices.

pgoyal01:
@iawells Thanks. We should make clear that this is only about control networks and we should address your statement "some means must be in place to get new software onto it".

user-stories/air-gapped-environments.md
```diff
@@ -0,0 +1,25 @@
...
+
+Note: Air gapped environments are not a challenge solely for service providers/network operators. Other operators of heavily regulated environments, such as US DoD networks, seeking to employ CNFs would be beholden to these constraints.
+
+## A CNF utilizes a cloud based licensing model
+
+A CNF running on a K8s cluster is downloaded from a centralized repository. Upon instantiation, the CNF validates that it has the required license tied to a suite of desired functionality. If the license is valid, the functionality is enabled. If a valid license is not available, the desired behavior is often that the next available license in a pool is requested, or a new order is generated in real time.
+
+In this case, if the operator does not allow for a CNF to access a remote licensing server, this method breaks down. Additionally, if the licensing server is pulled into the locally hosted environments but requires licenses be purchased in advance, then the desired behavior of dynamically deployed CNFs might be hindered due to a lack of available licenses in the pool.
```
electrocucaracha:
```suggestion
In this case, if the operator does not allow for a CNF to access a remote licensing server, this method breaks down. Additionally, if the licensing server is pulled into the locally hosted environments but requires licenses be purchased in advance, then the desired behavior of dynamically deployed CNFs might be hindered due to a lack of available licenses in the pool.
```

user-stories/air-gapped-environments.md
```diff
@@ -0,0 +1,25 @@
...
+
+The goal for licensing best practices should be to find ways that allow operators to safely and securely deploy CNFs on-demand in air gapped environments. This means that CNF vendors and operators must find a way to monetize CNFs that support cloud-based deployments in an environment that may break certain cloud native assumptions.
+
+## An operator employs [GitOps](https://github.com/cncf/cnf-wg/blob/main/use-case/0001-UC-lifecycle-of-infrastructure-where-CNF-is-running.md) methodologies with defense in depth
+
+In an air gapped environment, unique supply chain considerations arise. As discussed in [user-story supply-chain-attacks](https://github.com/cncf/cnf-wg/blob/main/user-stories/supply-chain-attacks.md), securing ones supply chain is paramount to a defense in depth strategy. In an operator's air gapped environment, this supply chain almost always enforces a centralized choke point for pulling, distributing and versioning artifacts.
```
electrocucaracha:
```suggestion
In an air gapped environment, unique supply chain considerations arise. As discussed in [user-story supply-chain-attacks](https://github.com/cncf/cnf-wg/blob/main/user-stories/supply-chain-attacks.md), securing ones supply chain is paramount to a defense in depth strategy. In an operator's air gapped environment, this supply chain almost always enforces a centralized choke point for pulling, distributing and versioning artifacts.
```

user-stories/air-gapped-environments.md
```diff
@@ -0,0 +1,25 @@
...
+
+In this case, the operator would seek to ensure that artifacts are signed and deemed "trusted" before being released to the rest of the interal actors. Trust could be established in multiple ways, such as internal pipelines that pull untrusted artifacts and run vulnerability scans and automated testing or via an established software supply chain with a trusted vendor. All declarations supporting an operator's GitOps style approach would strictly build from these privately hosted repositories in fully air gapped development, stage, and production environments. Testing would include validation that a CNF is fully deployable and maintainable with zero reach to the internet.
+
+## A vendor employs SaaS based services
+
+A vendor provides a SaaS based solution hosted in a third-party cloud environment. This service could be related to licensing, streaming telemetry, or automated fault detection. In this instance, the provider and operator must come to an agreement as to how these services will be accessed.
```
electrocucaracha:
```suggestion
A vendor provides a SaaS based solution hosted in a third-party cloud environment. This service could be related to licensing, streaming telemetry, or automated fault detection. In this instance, the provider and operator must come to an agreement as to how these services will be accessed.
```

electrocucaracha (commented):
Some linting issues regarding trailing spaces

user-stories/air-gapped-environments.md
```diff
@@ -0,0 +1,25 @@
+# User stories: Employing Air Gapped Environments
+
+Environments hosting critical infrastructure often deny access to the public internet, even via a proxy. This is done for several reasons such as controlling the artifacts that come into an environment, controlling builds at the source, and ensuring malicious code is not able to call back to remote destinations. Beyond the security implications of air gapping an environment and utilizing private registries, upstream repository management becomes a centralized task that all internal stakeholders can take advantage of. Internal developers and operations teams can rely on their SREs to ensure that the images, packages, etc... they require are available and secure.
```
electrocucaracha, line 3: IMHO, an air gapped environment short definition needs to be included here.

user-stories/air-gapped-environments.md
electrocucaracha: nit: relative paths
```suggestion
In an air gapped environment, unique supply chain considerations arise. As discussed in [user-story supply-chain-attacks](supply-chain-attacks.md), securing ones supply chain is paramount to a defense in depth strategy. In an operator's air gapped environment, this supply chain almost always enforces a centralized choke point for pulling, distributing and versioning artifacts. 
```

user-stories/air-gapped-environments.md
```diff
@@ -0,0 +1,25 @@
...
+
+In an air gapped environment, unique supply chain considerations arise. As discussed in [user-story supply-chain-attacks](https://github.com/cncf/cnf-wg/blob/main/user-stories/supply-chain-attacks.md), securing ones supply chain is paramount to a defense in depth strategy. In an operator's air gapped environment, this supply chain almost always enforces a centralized choke point for pulling, distributing and versioning artifacts. 
+
+In this case, the operator would seek to ensure that artifacts are signed and deemed "trusted" before being released to the rest of the interal actors. Trust could be established in multiple ways, such as internal pipelines that pull untrusted artifacts and run vulnerability scans and automated testing or via an established software supply chain with a trusted vendor. All declarations supporting an operator's GitOps style approach would strictly build from these privately hosted repositories in fully air gapped development, stage, and production environments. Testing would include validation that a CNF is fully deployable and maintainable with zero reach to the internet.
```
electrocucaracha:
```suggestion
<!---
TODO: Include GitOps definition reference once it's created (https://github.com/cncf/glossary/issues/265)
-->
In this case, the operator would seek to ensure that artifacts are signed and deemed "trusted" before being released to the rest of the interal actors. Trust could be established in multiple ways, such as internal pipelines that pull untrusted artifacts and run vulnerability scans and automated testing or via an established software supply chain with a trusted vendor. All declarations supporting an operator's GitOps style approach would strictly build from these privately hosted repositories in fully air gapped development, stage, and production environments. Testing would include validation that a CNF is fully deployable and maintainable with zero reach to the internet.
```

electrocucaracha (commented):
Another question that I have about this user story is related to the process to renew a license, I assume this depends on the mechanism used by the operator or vendor isn't it?

jeffsaelens:
@electrocucaracha 100% on the licensing. Proposing best practices on better licensing practices is something I'd like to spawn from this. It's a topic that I talk about with others frequently.

user-stories/air-gapped-environments.md
```diff
@@ -0,0 +1,25 @@
+# User stories: Employing Air Gapped Environments
```
taylor, line 1:
```suggestion
# User stories: Employing Air Gapped PLACES 
```

iawells: This sucks

user-stories/air-gapped-environments.md
```diff
@@ -0,0 +1,25 @@
...
+
+The goal for licensing best practices should be to find ways that allow operators to safely and securely deploy CNFs on-demand in air gapped environments. This means that CNF vendors and operators must find a way to monetize CNFs that support cloud-based deployments in an environment that may break certain cloud native assumptions.
+
+## An operator employs [GitOps](https://github.com/cncf/cnf-wg/blob/main/use-case/0001-UC-lifecycle-of-infrastructure-where-CNF-is-running.md) methodologies with defense in depth
+
+In an air gapped environment, unique supply chain considerations arise. As discussed in [user-story supply-chain-attacks](https://github.com/cncf/cnf-wg/blob/main/user-stories/supply-chain-attacks.md), securing ones supply chain is paramount to a defense in depth strategy. In an operator's air gapped environment, this supply chain almost always enforces a centralized choke point for pulling, distributing and versioning artifacts.
```
pgoyal01: typo
```suggestion
In an air gapped environment, unique supply chain considerations arise. As discussed in [user-story supply-chain-attacks](https://github.com/cncf/cnf-wg/blob/main/user-stories/supply-chain-attacks.md), securing one's supply chain is paramount to a defense in depth strategy. In an operator's air gapped environment, this supply chain almost always enforces a centralized choke point for pulling, distributing, and versioning artifacts.
```

user-stories/air-gapped-environments.md
```diff
@@ -0,0 +1,25 @@
...
+
+In this case, the operator would seek to ensure that artifacts are signed and deemed "trusted" before being released to the rest of the interal actors. Trust could be established in multiple ways, such as internal pipelines that pull untrusted artifacts and run vulnerability scans and automated testing or via an established software supply chain with a trusted vendor. All declarations supporting an operator's GitOps style approach would strictly build from these privately hosted repositories in fully air gapped development, stage, and production environments. Testing would include validation that a CNF is fully deployable and maintainable with zero reach to the internet.
+
+## A vendor employs SaaS based services
+
+A vendor provides a SaaS based solution hosted in a third-party cloud environment. This service could be related to licensing, streaming telemetry, or automated fault detection. In this instance, the provider and operator must come to an agreement as to how these services will be accessed.
```
pgoyal01, line 23: Is it enough to address only the access issue? What about how the SaaS service operates -- for example, does it interact with other services or has backdoors?

electrocucaracha:
I just recently discovered this [zarf tool](https://github.com/defenseunicorns/zarf) so I'm wondering if this can help on this user story.

iawells:
> I just recently discovered this [zarf tool](https://github.com/defenseunicorns/zarf) so I'm wondering if this can help on this user story.

It seems to do a little bit of everything but I'm interested in the app-package format.  (I'm less interested in the 'must run as root for no reason' part, but I think that's more about the k8s cluster creation side of things.)

user-stories/air-gapped-environments.md
```diff
@@ -0,0 +1,28 @@
+# User stories: Employing Air Gapped Environments
+
+Environments hosting critical infrastructure often deny access to the public internet, even via a proxy. This is done for several reasons such as controlling the artifacts that come into an environment, controlling builds at the source, and ensuring malicious code is not able to call back to remote destinations. Beyond the security implications of air gapping an environment and utilizing private registries, upstream repository management becomes a centralized task that all internal stakeholders can take advantage of. Internal developers and operations teams can rely on their SREs to ensure that the images, packages, etc... they require are available and secure.
```
pgoyal01, line 3: "controlling builds at the source" -- what is the "source"? In air-gapped environments, an operator would prefer to build locally, if that is possible (which would be rarely).

"upstream registry management" -- what is the "upstream"?  Ideally, following secure practices, including scanning and testing, items moved into a repository hosted in an air-gapped environment.

user-stories/air-gapped-environments.md
```diff
@@ -0,0 +1,28 @@
...
+
+## A CNF utilizes a cloud based licensing model
+
+A CNF running on a K8s cluster is downloaded from a centralized repository. Upon instantiation, the CNF validates that it has the required license tied to a suite of desired functionality. If the license is valid, the functionality is enabled. If a valid license is not available, the desired behavior is often that the next available license in a pool is requested, or a new order is generated in real time.
+
+In this case, if the operator does not allow for a CNF to access a remote licensing server, this method breaks down. Additionally, if the licensing server is pulled into the locally hosted environments but requires licenses be purchased in advance, then the desired behavior of dynamically deployed CNFs might be hindered due to a lack of available licenses in the pool.
+
+The goal for licensing best practices should be to find ways that allow operators to safely and securely deploy CNFs on-demand in air gapped environments. This means that CNF vendors and operators must find a way to monetize CNFs that support cloud-based deployments in an environment that may break certain cloud native assumptions.
```
pgoyal01: suggestion
```suggestion
The goal for licensing best practices should be to find ways that allow operators to safely and securely deploy CNFs on-demand in air gapped environments.
```

user-stories/air-gapped-environments.md
```diff
@@ -0,0 +1,28 @@
...
+
+## An operator employs [GitOps](https://github.com/cncf/cnf-wg/blob/main/use-case/0001-UC-lifecycle-of-infrastructure-where-CNF-is-running.md) methodologies with defense in depth
+
+In an air gapped environment, unique supply chain considerations arise. As discussed in [user-story supply-chain-attacks](https://github.com/cncf/cnf-wg/blob/main/user-stories/supply-chain-attacks.md), securing one's supply chain is paramount to a defense in depth strategy. In an operator's air gapped environment, this supply chain almost always enforces a centralized choke point for pulling, distributing, and versioning artifacts.
```
pgoyal01, line 17: Why is there a single choke point? The process to pull artifacts does not occur in an air-gapped environment. Similarly for the processes to scan, validate, etc.

user-stories/air-gapped-environments.md
```diff
@@ -0,0 +1,28 @@
...
+
+<!---
+TODO: Include GitOps definition reference once it's created (https://github.com/cncf/glossary/issues/265)
+-->
+In this case, the operator would seek to ensure that artifacts are signed and deemed "trusted" before being released to the rest of the interal actors. Trust could be established in multiple ways, such as internal pipelines that pull untrusted artifacts and run vulnerability scans and automated testing or via an established software supply chain with a trusted vendor. All declarations supporting an operator's GitOps style approach would strictly build from these privately hosted repositories in fully air gapped development, stage, and production environments. Testing would include validation that a CNF is fully deployable and maintainable with zero reach to the internet.
```
pgoyal01: I think security, even in non-air-gapped environments, requires that the "artifacts are signed". If the user story requires that the artifacts be signed by an operator's internal groups then let us state that.

Even "trusted" artifacts that are externally sourced need to be validated.

Is there no possibility of downloading a trusted image from a trusted vendor that has set up a secure site, and then subjecting it to certain validations prior to uploading to a repository in an air-gapped environment?

The validation testing is not only for deployment, maintenance, and zero-reach to the internet but also that it performs as intended and doesn't cause "internal damage".

```suggestion
In this case, the operator would seek to ensure that artifacts are signed and deemed "trusted" before being released to the rest of the internal actors. Trust could be established in multiple ways, such as internal pipelines that pull untrusted artifacts and run vulnerability scans and automated testing or via an established software supply chain with a trusted vendor. All declarations supporting an operator's GitOps style approach would strictly build from these privately hosted repositories in fully air gapped development, stage, and production environments. Testing would include validation that a CNF is fully deployable and maintainable with zero reach to the internet.
```

pgoyal01:
do we have a template for writing user stories? such as with a sho, what, why, etc.

user-stories/air-gapped-environments.md
electrocucaracha: > IMHO, an air gapped environment short definition needs to be included here.

Maybe this definition can be included in the CNCF glossary project.

user-stories/air-gapped-environments.md
jeffsaelens: Source is defined by the operator. This is their "source of truth that they have defined" via their internal repos. This is where they effect their changes and feed into their pipelines. They would "build locally" but they'll likely doing it from stuff drafted internally, and things they've pulled from elsewhere.

Upstream is anything north of that source of truth they have defined internally. This could be opensource repos, dockerhub, a vendor repository, etc.... This is where they are initially pulling images, charts, binaries and the like from into their air-gapped environments.

I think discussing this live would help.

user-stories/air-gapped-environments.md
jeffsaelens: You have to control how things get into the environment. So, you create an ingress for artifacts destined for air-gapped production environments by controlling how/where these artifacts get introduced.

user-stories/air-gapped-environments.md
jeffsaelens: I'll take this as an action item (unless this has already been assigned during a call). Playing some catchup after the recent career move.

user-stories/air-gapped-environments.md
pgoyal01: @jeffsaelens Thanks. Can we define the terms as you have specified

user-stories/use-caseair-gapped-environments.md
pgoyal01: @jeffsaelens Thanks - can you reword?

user-stories/air-gapped-environments.md
jeffsaelens: @pgoyal01 and @electrocucaracha would it be more helpful for me define these terms in their sub-sections, or to create a mini-glossary at the bottom? As of now, I'm tracking I need to provide contextual definitions for:

- air-gapped
- Source (from the air-gapped/providers POV)
- Upstream (from the air-gapped/providers POV)

user-stories/air-gapped-environments.md
jeffsaelens: Too wordy with the contextual follow-up?

user-stories/air-gapped-environments.md
jeffsaelens: This is where we'll expand into best practices. VPNs, Proxies, and the like may be required to enact this. So how does a CSP do this smartly? CSPs are going to have to give in some instances if they want access to all the cool tech. Where does it make sense to capitulate and where are you assuming too much risk? Conversely, the best practice might be to do away with air-gapped environments and build out real ZT environments. (doing this at scale in the near term though seems about as likely as finding a unicorn imho however)

Ironically, we could go down a whole rabbit hole in implementing a cloud native proxy (which in this case would be a very entry level CNF potentially).

user-stories/air-gapped-environments.md
iawells:
```suggestion
The goal for licensing best practices should be to find ways that allow operators to safely and securely deploy CNFs on-demand in air gapped environments. This means that CNF vendors and operators must find a way to operate CNFs in this environment - one that breaks a common assumption of cloud software, that it has reachability to internet services.
```

user-stories/air-gapped-environments.md
iawells: The cloud must receive its software from somewhere, conventionally by pulling (though other options like prepopulation are available).  We could divide this into problem and action: one is that you can't pull from on-internet repositories, the other is that the choke point that @jeffsaelens suggests is actually the best alternative.

user-stories/air-gapped-environments.md
```diff
@@ -0,0 +1,31 @@
...
+
+This could be accomplished via a VPN, by providing operators the ability to host these software solutions within their environments, or convincing operators that fully air gapped approaches are not feasible in the cloud native world. Best practices that explore how to consume SaaS based offers in air gapped environments must be explored in the future.
+
+## Glossary
+
+- Air gapped environment: A physically and logically isolated environment that disallows direct network access/connectivity. These environments are created to protect infrastructure from bad actors either attempting to gain access to the environment remotely, or for malicious software to reach outside of the isolated environment. Given these constraints, standard cloud software consumption models where users and software pull, clone, or download required resources from public repositories are blocked. All new software bound for this environment must first be introduced through the operator’s secure source control mechanisms.
```
iawells, line 29: Perhaps introduce the Platonic ideal - there is no physical connectivity - and mention that there are ways short of this that are not truly airgapped but might be acceptable compromises with fewer constraints.

iawells: Perhaps introduce 'isolated environment' as a term that can mean both airgapped and the weaker forms?

user-stories/air-gapped-environments.md
```diff
@@ -0,0 +1,31 @@
...
+- Upstream: For the context of this user story, upstream refers to any repository sitting "North" of the private repository. Air gapped environments, if built to be truly isolated, are prevented from pulling directly from upstream repositories. To access upstream software in an air-gapped environment, software should be validated and then hosted in a private repository. Mechanisms, such as VPNs and proxies can be leveraged as part of a secure source control strategy as well, but implies implicit trust to the additional “sources” you are pulling from.
```
iawells:
```suggestion
- Upstream: For the context of this user story, upstream refers to any repository sitting "North" of the private repository. Air gapped environments, if built to be truly isolated, are prevented from pulling directly from internet repositories and cannot use them as the upstream repository. To access software in an air-gapped environment, software should be validated and then hosted in a private repository within the airgapped environment. Mechanisms, such as VPNs and proxies can be leveraged as part of a secure source control strategy as well, but implies implicit trust to the additional “sources” you are pulling from and the security of the network mechanism used.
```

user-stories/air-gapped-environments.md
pgoyal01: @jeffsaelens Sorry for the delay in responding. A mini-glossary would be easier for folks, I think.

taylor (approved):
lgtm

Smitholi67 (approved):
/LGTM

lixuna (approved):
LGTM

Issue #242, opened 2022-12-08
Best practice proposal: A CNFs containers should have one process category (or type)


taylor:
>A microservice should have only one process (or set of parent/child processes) that is managed by a non home grown supervisor or orchestrator. The microservice should not spawn other process types (e.g. executables) as a way to contributeto the workload but rather should interact with other processes through a microservice API.
from [CNF Test Suite](https://github.com/cncf/cnf-testsuite/blob/main/RATIONALE.md#to-check-if-the-cnf-has-multiple-process-types-within-one-container-single_process_type)

[WIP / draft proposal](https://docs.google.com/document/d/17FeO3zbGsf6pzXGXTd_59eTo8bZdBCtImCxFubr5Zzs/edit)

Reference(s):
- [Docker on decoupling applications🔗](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#decouple-applications)
- CNF Certification test: https://github.com/cncf/cnf-certification/blob/main/docs/CNFCertification-1.0-beta.md#single-process-type-in-one-container

taylor:
Feedback requested on this as a best practice @iawells @electrocucaracha @wavell @denverwilliams @fkautz @michaelspedersen @CsatariGergely @ildikov

taylor:
@tomkivlin

tomkivlin:
Link to rationale in cnf-test-suite: https://github.com/cncf/cnf-testsuite/blob/main/RATIONALE.md#to-check-if-the-cnf-has-multiple-process-types-within-one-container-single_process_type

CsatariGergely:
I think this should _not_ be a best practice and it should not be an essential test in CNF Conformance. Complex applications, like telco ones have several situations where more than one process is needed by a microservice as there are lots of networking and storage things happen while we are in a time constrained environment. Organizing each process type to a separate container leads to scaling and dimensioning issues and waste of resources. 
Scaling of processes is different than containers, processes are lifecycle managed and monitored by the Linux kernel while containers are handled by the container runtime including some feedback to controller via kubelet what makes containers less scalable than processes.
When we have each process type in separate containers we need to dimension each and every of these separately what eventually leads small resource requests for containers and invariability in container performance. To overcome this resource requests are increased and the utilization and performance of the whole CNF will be low. 
Containers are just more heavy than processes, so this practice results in waste of resources. 
All in all I do not see this best practice to be justified and I think we should not have it. 
[...]

PR #267, opened 2023-09-07
Create 0005-single-concern-per-container.md

taylor:
Adding the initial version of best practice recommending using a single concern (or single process type) per container

A CNF’s microservice(s) should follow the single concern principle. To help achieve this goal, a CNF’s microservice(s) should have only one process type (or set of parent/child processes) per container. The process(es) in the container should not spawn other process types (e.g. executables) as a way to contribute to the workload but rather should interact with other processes through a microservice API.

Co-authored with: @Smitholi67 and @electrocucaracha

Issue: #242

### DIFF
--- a/doc/cbpps/0005-single-concern-per-container.md
+++ b/doc/cbpps/0005-single-concern-per-container.md
@@ -1,0 +1,176 @@
+# **CBPP-0005: A CNF’s containers should have one process category**
+
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+- [Workload Context](#workload-context)
+  - [User Stories (Optional)](#user-stories)
+  - [Notes/Constraints/Caveats (Optional)](#notesconstraintscaveats-optional)
+  - [References](#references)
+- [Test Plan](#testing-objectives)
+- [Implementation History](#implementation-history)
+
+## **Release Signoff Checklist**
+
+Items marked with (R) are required for the proposed best practice to be included in a release.
+
+- [ ] (R) CBPP approvers have approved the CBPP status as `implementable`
+- [ ] (R) CBPP summary, motivation and best practice details are appropriately documented
+- [ ] (R) Test plan is in place, giving consideration to CNF Test Suite input
+- [ ] (R) Scoring has been determined
+- [ ]   "Implementation History" section is up-to-date
+- [ ]    Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+## **Summary**
+
+A CNF’s microservice(s) should follow the single concern principle. To help achieve this goal, a CNF’s microservice(s) should have only one process type (or set of parent/child processes) per container. The process(es) in the container should not spawn other process types (e.g. executables) as a way to contribute to the workload but rather should interact with other processes through a microservice API.
+
+_“It’s best practice to separate areas of concern by using one service per container. That service may fork into multiple processes (for example, the Apache web server starts multiple worker processes). It’s ok to have multiple processes, but to get the most benefit out of Docker, avoid one container being responsible for multiple aspects of your overall application.”_
+	&nbsp;&nbsp;&nbsp;ref: https://docs.docker.com/config/containers/multi-service_container/
+
+
+This will help the programmability and flexibility of CNFs and networks through:
+- Improvements in scalability and responsiveness to service needs
+- Offloading container orchestration and leveraging the platform's capabilities
+- Reducing downtime risk in deployments and upgrades, increasing the speed of delivery of services
+- Managing service concerns in independent units
+
+
+## **Motivation**
+
+Challenges we think this best practice can help solve:
+
+- Life-cycle management
+    - If you have more than one process type in a container, you will have to manage the lifecycle of the secondary concerns within the container, which effectively means creating an orchestrator within the parent process type, reducing the value of using the Kubernetes orchestration capabilities.
+    - Resource utilization is less efficient in multi-concerned containers, which require allocation for all components rather than individual microservices requirements.
+    - Response time of multi-concerned containers is increased as scaling is required for the combined services rather than the individual services needing scaling.
+    - Upgrades causing interruption in many services and all their integration at once because none of the components in the system are independent 
+- Security
+    - CNFs with multi-concerned containers have a larger surface area for security attacks and production bugs
+    - Security vulnerabilities in one process type effect all other processes in the same container
+- Observability
+    - Reduced visibility of communication and activity of services in the multi-concerned container
+    - Log messages from the multi-concerned container are more complex because they are from many different sources instead of a single process type
+    - Difficulty in identifying which process a bug resides in increases as a result of multiple process types in a single container
+- Software Development Cycle
+    - A CNF with multiple maintainers (which can be different groups in an organization) can block each other’s development process when a CNF is tightly coupled
+    - A high degree of test coverage with a complex and tightly coupled CNF is difficult to achieve
+
+
+### **Goals**
+
+- Life-cycle management
+    - Increase the use of Kubernetes orchestration capabilities instead of using additional container orchestration solutions 
+    - Align with microservice architectural practices for operations and development
+    - Make it easier to scale multiple containers of a CNF to produce an efficient resource utilization and faster response:
+        - Make it easier to reason the best way to scale a CNF based on anticipated load by reducing the complexity of service concerns and allowing focus on each container (or Pods) anticipated needs 
+        - Support scaling individual processes of the same type for efficient resource utilization and faster response time for changes in service load (increasing/decreasing)
+    - Simplify deployment, reduce risk in upgrades and support easier rollbacks by managing the process types (service concerns) independently and providing coarse-grained dependencies at the container level 
+- Security
+    - Reducing attack surface area in a CNF’s containers by limiting the number process types and their dependencies such as additional binaries/libraries 
+    - Protect containers from interfering with one another by leveraging the container namespace system
+    - Allow finer control of permissions
+- Observability
+    - Simplifying troubleshooting and readability of log output. It is easier to reason about the log messages that come from one concern or process type than log messages that are interleaved with other concerns. This is even more true in a container that prints all log messages to standard out, such as in a 12-factor cloud native app.
+    - Improve ability to monitor CNF activity by exposing the inter-process communication
+- Software Development Cycle
+    - Enable better support for multiple development teams to maintain their own independent lifecycle(s)
+    - Move towards loosely coupled components: 
+        - Reduces the complexity of each individual component
+        - Simplifies determining where code resides for a given piece of functionality
+        - Increases the confidence in test coverage of individual components
+        - Facilitates locating the root cause of problems by eliminating extra variables
+        - Minimizes the risk of changes in one component causing problems with another component. 
+    - Increase composability of the CNF and it’s component services
+        - One concern, and therefore process type, per container is much easier to reason about because it is easier to share and verbally communicate about its contents digitally. This makes it easier to reuse in other projects
+        - Loosely coupled containers with well defined service APIs increase interoperability with other CNFs as well as between a CNFs component services
+
+### **Non-Goals**
+
+- Best practices for CNFs communication with other processes through microservice APIs
+- Best practices on process supervision and management
+- Reduce risk issues from of home-grown process management systems
+    - Security, Zombie processes, unknown failures
+- Specifying implementation details of each component
+- Specifying how a CNF should be split into individual micro services
+
+## **Proposal**
+
+A CNF with multiple concerns should split services (or process types) for each of its concerns into separate containers. Service dependencies should be handled between containers through well-defined interfaces. 
+
+Pod specs for the CNF should provide scaling and monitoring information for each service running in different containers.
+
+## **Workload Context**
+
+All pod types should implement this best practice.
+
+### **User Stories**
+
+#### 5G applications needing high-degree of flexibility and automation
+
+_“5G requirements range from enhanced mobile broadband to ultra-reliable, low-latency communications to massive machine-type communications. Various 5G applications will mix and match these characteristics, requiring a high degree of programmability and the ability to easily combine different functionalities.”_  from “Why Use Containers and CloudNative Functions Anyway?” by Muthurajan Jayakumar (M Jay) @ Intel 
+
+Separating service concerns, represented by process types, into different containers supports the goal of flexibility and programmability allowing the developer to provide definitions for the services needs, including hardware requirements, and the CNF consumer to efficiently scale these 5G services while supporting the accommodation of any container specific needs. 
+
+Separating service concerns also helps to support the automation goals of the 5G operators by providing a more composable CNF with well defined interfaces and coarse grained dependencies at the container level increasing confidence in the testing phases of their pipelines and lowering risk during upgrades. 
+
+
+#### SMF with multiple services for communicating with the AMF, UPF, and PCF needing more responsive scaling and requiring more resources for its AMF and UPF communication
+
+An SMF with different services providing the communication between AMF, UPF, PCF and other services in an environment with many new sessions being initiated sessions and ending during peak times may require scaling of the service responsible for communication with the AMF and UPF faster than the service communicating with the PCF. If the services are split into their own containers they can be scaled independently.  
+
+
+
+
+![Session Management Function within a 5G Service-based Architecture](https://hackmd.io/_uploads/B1qwravAh.png)
+_Source: https://techcommunity.microsoft.com/t5/azure-for-operators-blog/what-is-the-5g-session-management-function-smf/ba-p/3693852_
+
+
+
+### **Notes/Constraints/Caveats**
+
+“That service may fork into multiple processes (for example, the Apache web server starts multiple worker processes). It’s ok to have multiple processes, but to get the most benefit out of Docker, avoid one container being responsible for multiple aspects of your overall application.”
+
+
+Ideally in a container the process 1 (init process) should be the single application running in the container. When multiple processes are started it is recommended to use a non-home-grown supervisor such as http://supervisord.org/.  See discussion here https://github.com/cncf/cnf-wg/discussions/264
+
+
+### Definitions
+ 
+
+- Monolithic application - applications which do not separate concerns into microservices are considered monolithic
+    - Monolithic CNF - a monolithic application focused on networking concerns
+- Multi-concerned containers - a container having more than a single process type providing services for different concerns
+
+
+
+### **References**
+
+- https://www.infoq.com/articles/cloud-native-network-functions-concern/
+- https://www.ibm.com/cloud/architecture/architecture/practices/cloud-native-principles
+    - _“SoC (separation of concerns) - Single concern principle”_
+- https://docs.docker.com/config/containers/multi-service_container/
+- https://en.wikipedia.org/wiki/Single-responsibility_principle
+- https://developers.redhat.com/articles/2022/01/11/5-design-principles-microservices
+- https://www.ericsson.com/en/future-technologies/architecture/network-capabilities
+_“As a step towards a fully autonomous network and achieving an intent-based management of a network, its architecture must be prepared by raising the level of abstraction in management with e.g., strong separation of concerns.”_
+- https://github.com/cncf/cnf-testsuite/blob/main/docs/LIST_OF_TESTS.md#single-process-type-in-one-container
+    - RATIONALE - https://github.com/cncf/cnf-testsuite/blob/main/RATIONALE.md#to-check-if-the-cnf-has-multiple-process-types-within-one-container-single_process_type
+- https://www.tutorialworks.com/containers-single-or-multiple-processes/
+- https://www.intel.com/content/dam/www/public/us/en/documents/white-papers/containers-and-cloud-native-functions-white-paper.pdf
+- https://techcommunity.microsoft.com/t5/azure-for-operators-blog/what-is-the-5g-session-management-function-smf/ba-p/3693852
+
+
+## **Testing Objectives**
+
+Validate if there are more than one process type running inside of a container
+
+## **Implementation History**
+
+Documentation:
+  - First version: September 2023
+Test: CNF Test Suite
+  - First version: [July 2021](https://github.com/cncf/cnf-testsuite/commit/7dc8934e1f9bf643f503ca695f6685cb22b18975)
--- a/doc/cbpps/0005-single-concern-per-container.md
+++ b/doc/cbpps/0005-single-concern-per-container.md
@@ -62,7 +62,7 @@ Challenges we think this best practice can help solve:
 ### **Goals**
 
 - Life-cycle management
-    - Increase the use of Kubernetes orchestration capabilities instead of using additional container orchestration solutions 
+    - Simplify and consolidate the use of orchestration capabilities instead of adding additional container orchestration solutions 
     - Align with microservice architectural practices for operations and development
     - Make it easier to scale multiple containers of a CNF to produce an efficient resource utilization and faster response:
         - Make it easier to reason the best way to scale a CNF based on anticipated load by reducing the complexity of service concerns and allowing focus on each container's anticipated needs 
--- a/doc/cbpps/0005-single-concern-per-container.md
+++ b/doc/cbpps/0005-single-concern-per-container.md
@@ -45,7 +45,7 @@ Challenges we think this best practice can help solve:
 
 - Life-cycle management
     - If you have more than one process type in a container, you will have to manage the lifecycle of the secondary concerns within the container, which effectively means creating an orchestrator within the parent process type, reducing the value of using the Kubernetes orchestration capabilities.
-    - Resource utilization is less efficient in multi-concerned containers, which require allocation for all components rather than individual microservices requirements.
+    - Resource utilization is likely to be less efficient in multi-concerned containers, as the container runtime will look to allocation infrastructure resources for all components rather than individual microservices requirements.
     - Response time of multi-concerned containers is increased as scaling is required for the combined services rather than the individual services needing scaling.
     - Upgrades causing interruption in many services and all their integration at once because none of the components in the system are independent 
 - Security
@@ -120,7 +120,7 @@ Separating service concerns also helps to support the automation goals of the 5G
 
 #### SMF with multiple services for communicating with the AMF, UPF, and PCF needing more responsive scaling and requiring more resources for its AMF and UPF communication
 
-An SMF with different services providing the communication between AMF, UPF, PCF and other services in an environment with many new sessions being initiated sessions and ending during peak times may require scaling of the service responsible for communication with the AMF and UPF faster than the service communicating with the PCF. If the services are split into their own containers they can be scaled independently.  
+An SMF with different services providing the communication between AMF, UPF, PCF and other services in an environment where many new sessions are initiated and end during peak times. This may require scaling of the service responsible for communication with the AMF and UPF faster than the service communicating with the PCF. If the services are split into their own containers they can be scaled independently.  
 
 
 
@@ -135,7 +135,7 @@ _Source: https://techcommunity.microsoft.com/t5/azure-for-operators-blog/what-is
 “That service may fork into multiple processes (for example, the Apache web server starts multiple worker processes). It’s ok to have multiple processes, but to get the most benefit out of Docker, avoid one container being responsible for multiple aspects of your overall application.”
 
 
-Ideally in a container the process 1 (init process) should be the single application running in the container. When multiple processes are started it is recommended to use a non-home-grown supervisor such as http://supervisord.org/.  See discussion here https://github.com/cncf/cnf-wg/discussions/264
+Ideally in a container the process 1 (init process) should be the single application running in the container, as the container runtime monitors the PID 1 application process and uses its signals to report events, knowing when a container has stopped. When multiple processes are started it is recommended to use a non-home-grown supervisor such as http://supervisord.org/.  See discussion here https://github.com/cncf/cnf-wg/discussions/264
 
 
 ### Definitions
--- a/doc/cbpps/0005-single-concern-per-container.md
+++ b/doc/cbpps/0005-single-concern-per-container.md
@@ -163,6 +163,7 @@ _“As a step towards a fully autonomous network and achieving an intent-based m
 1. https://www.tutorialworks.com/containers-single-or-multiple-processes/
 1. https://www.intel.com/content/dam/www/public/us/en/documents/white-papers/containers-and-cloud-native-functions-white-paper.pdf
 1. https://techcommunity.microsoft.com/t5/azure-for-operators-blog/what-is-the-5g-session-management-function-smf/ba-p/3693852
+1. https://www.mirantis.com/blog/multi-container-pods-and-container-communication-in-kubernetes/
 
 
 ## **Testing Objectives**
--- a/doc/cbpps/0005-single-concern-per-container.md
+++ b/doc/cbpps/0005-single-concern-per-container.md
@@ -65,7 +65,7 @@ Challenges we think this best practice can help solve:
     - Increase the use of Kubernetes orchestration capabilities instead of using additional container orchestration solutions 
     - Align with microservice architectural practices for operations and development
     - Make it easier to scale multiple containers of a CNF to produce an efficient resource utilization and faster response:
-        - Make it easier to reason the best way to scale a CNF based on anticipated load by reducing the complexity of service concerns and allowing focus on each container (or Pods) anticipated needs 
+        - Make it easier to reason the best way to scale a CNF based on anticipated load by reducing the complexity of service concerns and allowing focus on each container's anticipated needs 
         - Support scaling individual processes of the same type for efficient resource utilization and faster response time for changes in service load (increasing/decreasing)
     - Simplify deployment, reduce risk in upgrades and support easier rollbacks by managing the process types (service concerns) independently and providing coarse-grained dependencies at the container level 
 - Security
--- a/doc/cbpps/0005-single-concern-per-container.md
+++ b/doc/cbpps/0005-single-concern-per-container.md
@@ -70,15 +70,15 @@ Challenges we think this best practice can help solve:
 - Security
   - Increase confidentiality by creating clearly defined boundaries between software components
     - Reduces the attack surface presented in a CNF’s containers
-    - Limit the number of process types and their dependencies such as additional binaries/libraries  
-    - Prevent unnecessary access to data via shared filesystems
-    - Implement finer grained attribute-based controls between processes and systems
+    - Limits the number of process types and their dependencies such as additional binaries/libraries  
+    - Prevents unnecessary access to data via shared filesystems
+    - Allows for finer grained attribute-based controls to be implemented between processes and systems
   - Increase integrity by restricting a processes ability to perform CRUD operations in a shared container environment
-    - Protect containers from interfering with one another by leveraging the container namespace system and cgroup implementations
+    - Protects containers from interfering with one another by leveraging the container namespace system and cgroup implementations
   - Increase availability by reducing the size of failure domain
     - Allows for granular control of separation of concerns
-    - Remove the risk of multiple critical processes terminating simultaneously
-    - Define access methods via established interfaces
+    - Removes the risk of multiple critical processes terminating simultaneously
+    - Defines access methods via established interfaces
 - Observability
   - Simplifying troubleshooting and readability of log output. It is easier to consume log messages and reason about their output when they come from one concern or process type as opposed to when they are interweaved with other concerns. This is even more true in a container that prints all log messages to standard out, such as described in the 12-factor cloud native app.
   - Improve ability to monitor CNF activity by exposing the inter-process communication
--- a/doc/cbpps/0005-single-concern-per-container.md
+++ b/doc/cbpps/0005-single-concern-per-container.md
@@ -45,11 +45,11 @@ Challenges we think this best practice can help solve:
 
 - Life-cycle management
     - If you have more than one process type in a container, you will have to manage the lifecycle of the secondary concerns within the container, which effectively means creating an orchestrator within the parent process type, reducing the value of using the Kubernetes orchestration capabilities.
-    - Resource utilization is likely to be less efficient in multi-concerned containers, as the container runtime will look to allocation infrastructure resources for all components rather than individual microservices requirements.
+    - Resource utilization is likely to be less efficient in multi-concerned containers, as the container runtime will look to allocate infrastructure resources for all components rather than individual microservices requirements.
     - Response time of multi-concerned containers is increased as scaling is required for the combined services rather than the individual services needing scaling.
     - Upgrades causing interruption in many services and all their integration at once because none of the components in the system are independent 
 - Security
-    - CNFs with multi-concerned containers have a larger surface area for security attacks and production bugs
+    - CNFs with multi-concerned containers have a larger surface area for cyberattacks and production bugs
     - Security vulnerabilities in one process type affect all other processes in the same container
 - Observability
     - Reduced visibility of communication and activity of services in the multi-concerned container, as the container runtime will only be monitoring the init process within the container for signals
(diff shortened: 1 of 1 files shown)


doc/cbpps/0005-single-concern-per-container.md
```diff
@@ -0,0 +1,176 @@
...
+_Source: https://techcommunity.microsoft.com/t5/azure-for-operators-blog/what-is-the-5g-session-management-function-smf/ba-p/3693852_
+
+
+
+### **Notes/Constraints/Caveats**
+
+“That service may fork into multiple processes (for example, the Apache web server starts multiple worker processes). It’s ok to have multiple processes, but to get the most benefit out of Docker, avoid one container being responsible for multiple aspects of your overall application.”
+
+
+Ideally in a container the process 1 (init process) should be the single application running in the container. When multiple processes are started it is recommended to use a non-home-grown supervisor such as http://supervisord.org/.  See discussion here https://github.com/cncf/cnf-wg/discussions/264
```
tomkivlin: Perhaps add some of the additional context that was provided in the discussion, e.g.

```suggestion
Ideally in a container the process 1 (init process) should be the single application running in the container, as the container runtime monitors the PID 1 application process and uses its signals to report events, knowing when a container has stopped. When multiple processes are started it is recommended to use a non-home-grown supervisor such as http://supervisord.org/.  See discussion here https://github.com/cncf/cnf-wg/discussions/264
```

doc/cbpps/0005-single-concern-per-container.md
```diff
@@ -0,0 +1,176 @@
...
+
+Challenges we think this best practice can help solve:
+
+- Life-cycle management
+    - If you have more than one process type in a container, you will have to manage the lifecycle of the secondary concerns within the container, which effectively means creating an orchestrator within the parent process type, reducing the value of using the Kubernetes orchestration capabilities.
+    - Resource utilization is less efficient in multi-concerned containers, which require allocation for all components rather than individual microservices requirements.
+    - Response time of multi-concerned containers is increased as scaling is required for the combined services rather than the individual services needing scaling.
+    - Upgrades causing interruption in many services and all their integration at once because none of the components in the system are independent 
+- Security
+    - CNFs with multi-concerned containers have a larger surface area for security attacks and production bugs
```
tomkivlin: Should we add a "may" here, as it's not a given?

```suggestion
    - CNFs with multi-concerned containers may have a larger surface area for security attacks and production bugs
```

doc/cbpps/0005-single-concern-per-container.md
```diff
@@ -0,0 +1,176 @@
...
+    - Security vulnerabilities in one process type effect all other processes in the same container
+- Observability
+    - Reduced visibility of communication and activity of services in the multi-concerned container
```
tomkivlin: Suggest adding a "because" statement to aid understanding, e.g.

```suggestion
    - Reduced visibility of communication and activity of services in the multi-concerned container, as the container runtime will only be monitoring the init process within the container for signals
```

doc/cbpps/0005-single-concern-per-container.md
```diff
@@ -0,0 +1,176 @@
...
+Challenges we think this best practice can help solve:
+
+- Life-cycle management
+    - If you have more than one process type in a container, you will have to manage the lifecycle of the secondary concerns within the container, which effectively means creating an orchestrator within the parent process type, reducing the value of using the Kubernetes orchestration capabilities.
+    - Resource utilization is less efficient in multi-concerned containers, which require allocation for all components rather than individual microservices requirements.
+    - Response time of multi-concerned containers is increased as scaling is required for the combined services rather than the individual services needing scaling.
+    - Upgrades causing interruption in many services and all their integration at once because none of the components in the system are independent 
+- Security
+    - CNFs with multi-concerned containers have a larger surface area for security attacks and production bugs
+    - Security vulnerabilities in one process type effect all other processes in the same container
```
tomkivlin:
```suggestion
    - Security vulnerabilities in one process type are likely to effect all other processes in the same container
```

doc/cbpps/0005-single-concern-per-container.md
```diff
@@ -0,0 +1,176 @@
...
+- Managing service concerns in independent units
+
+
+## **Motivation**
+
+Challenges we think this best practice can help solve:
+
+- Life-cycle management
+    - If you have more than one process type in a container, you will have to manage the lifecycle of the secondary concerns within the container, which effectively means creating an orchestrator within the parent process type, reducing the value of using the Kubernetes orchestration capabilities.
+    - Resource utilization is less efficient in multi-concerned containers, which require allocation for all components rather than individual microservices requirements.
```
tomkivlin: This might be contentious (see https://github.com/cncf/cnf-wg/issues/242#issuecomment-1363755449).  I think we should soften the statement a little, e.g.

```suggestion
    - Resource utilization is likely to be less efficient in multi-concerned containers, as the container runtime will look to allocation infrastructure resources for all components rather than individual microservices requirements.
```

doc/cbpps/0005-single-concern-per-container.md
```diff
@@ -0,0 +1,176 @@
...
+
+Separating service concerns also helps to support the automation goals of the 5G operators by providing a more composable CNF with well defined interfaces and coarse grained dependencies at the container level increasing confidence in the testing phases of their pipelines and lowering risk during upgrades. 
+
+
+#### SMF with multiple services for communicating with the AMF, UPF, and PCF needing more responsive scaling and requiring more resources for its AMF and UPF communication
+
+An SMF with different services providing the communication between AMF, UPF, PCF and other services in an environment with many new sessions being initiated sessions and ending during peak times may require scaling of the service responsible for communication with the AMF and UPF faster than the service communicating with the PCF. If the services are split into their own containers they can be scaled independently.
```
tomkivlin: This is a long sentence and I'm not sure it reads right for me. Maybe the following?

```suggestion
An SMF with different services providing the communication between AMF, UPF, PCF and other services in an environment where many new sessions are initiated and end during peak times. This may require scaling of the service responsible for communication with the AMF and UPF faster than the service communicating with the PCF. If the services are split into their own containers they can be scaled independently.  
```

tomkivlin (commented):
Some minor readability / understanding comments.

doc/cbpps/0005-single-concern-per-container.md
taylor: No objections on cnf wg call.  accepting

doc/cbpps/0005-single-concern-per-container.md
taylor: @tomkivlin by definition two process types will increase the surface area for attacks and bugs vs a single process type in a container.

doc/cbpps/0005-single-concern-per-container.md
```diff
@@ -0,0 +1,176 @@
...
+- Security
+    - CNFs with multi-concerned containers have a larger surface area for security attacks and production bugs
+    - Security vulnerabilities in one process type effect all other processes in the same container
+- Observability
+    - Reduced visibility of communication and activity of services in the multi-concerned container
+    - Log messages from the multi-concerned container are more complex because they are from many different sources instead of a single process type
+    - Difficulty in identifying which process a bug resides in increases as a result of multiple process types in a single container
+- Software Development Cycle
+    - A CNF with multiple maintainers (which can be different groups in an organization) can block each other’s development process when a CNF is tightly coupled
+    - A high degree of test coverage with a complex and tightly coupled CNF is difficult to achieve
```
nickolaev: Enables "polyglot" designs where e.g. the fronted of a service can be written in NodeJS, while your backend is in Erlang!

doc/cbpps/0005-single-concern-per-container.md
taylor:
```suggestion
    - Security vulnerabilities in one process type may affect all other processes in the same container
```

taylor: @tomkivlin I think this should be "may effect" to make it lighter

lixuna:
```suggestion
    - Security vulnerabilities in one process type affect all other processes in the same container
```

doc/cbpps/0005-single-concern-per-container.md
taylor: @tomkivlin, @electrocucaracha and I think the 
> (unless you're using a custom supervisor)

is not needed. The custom supervisor would still limit general observability as well and may require custom logging and service exposure...  If we have it please add to the definitions section for the custom supervisor.

doc/cbpps/0005-single-concern-per-container.md
taylor: Accepted during cnf wg call

doc/cbpps/0005-single-concern-per-container.md
taylor: Accepted during cnf wg call?

doc/cbpps/0005-single-concern-per-container.md
taylor: @nickolaev please do a suggest edit, this is a good one.

doc/cbpps/0005-single-concern-per-container.md
```diff
@@ -0,0 +1,176 @@
...
+    - CNFs with multi-concerned containers have a larger surface area for security attacks and production bugs
+    - Security vulnerabilities in one process type affect all other processes in the same container
+- Observability
+    - Reduced visibility of communication and activity of services in the multi-concerned container
+    - Log messages from the multi-concerned container are more complex because they are from many different sources instead of a single process type
+    - Difficulty in identifying which process a bug resides in increases as a result of multiple process types in a single container
+- Software Development Cycle
+    - A CNF with multiple maintainers (which can be different groups in an organization) can block each other’s development process when a CNF is tightly coupled
+    - A high degree of test coverage with a complex and tightly coupled CNF is difficult to achieve
+
```
nickolaev:
```suggestion
    - Enable "polyglot" software designs, where the decoupled entities are developed in entirely different frameworks and even computer languages

```

doc/cbpps/0005-single-concern-per-container.md
```diff
@@ -0,0 +1,176 @@
...
+
+### **Notes/Constraints/Caveats**
+
+“That service may fork into multiple processes (for example, the Apache web server starts multiple worker processes). It’s ok to have multiple processes, but to get the most benefit out of Docker, avoid one container being responsible for multiple aspects of your overall application.”
+
+
+Ideally in a container the process 1 (init process) should be the single application running in the container, as the container runtime monitors the PID 1 application process and uses its signals to report events, knowing when a container has stopped. When multiple processes are started it is recommended to use a non-home-grown supervisor such as http://supervisord.org/.  See discussion here https://github.com/cncf/cnf-wg/discussions/264
+
+
```
nickolaev, line 141:
```suggestion
Depending on the specifics of the implemented functionality, the resulting split may have requirements towards the locality of the resulting multiple containers. It may require certain functions to be deployed on the same worker node or allow for distributing them amongst different worker nodes. The cloud-native principles, and the need for a horizontal scaling does imply the latter, i.e. no locality restrictions. However, due to the specifics of the Telco workloads, such limitations might be necessary for the operation of the particular functionality.
In some cases, the workload distribution can span multiple data-centers and geographic locations. The edge deployments are exemplifying such application designs.

```

doc/cbpps/0005-single-concern-per-container.md
taylor: agreed, lgtm

doc/cbpps/0005-single-concern-per-container.md
taylor: lgtm for notes section.  

@nickolaev this topic would be interesting to explore to see how to give recommendations for the workloads that should allow running on multiple nodes and the ones that can not or are more difficult to do so.  For the ones that can not be split we should look at alternative best practices to help them

doc/cbpps/0005-single-concern-per-container.md
tomkivlin: @taylor ok, understood. I've updated the suggestion accordingly.

doc/cbpps/0005-single-concern-per-container.md
tomkivlin: OK, happy to withdraw this one.

tomkivlin (approved):
lgtm once @nickolaev's changes are incorporated.

doc/cbpps/0005-single-concern-per-container.md
taylor: lgtm

nickolaev (approved):

CsatariGergely:
I think this is not a best practice for CNF-s. Restricting to single concern or process type per container results in too many API calls in the architecture what gives a too big overhead to the latency sensitive CNF-s.

doc/cbpps/0005-single-concern-per-container.md
```diff
@@ -0,0 +1,179 @@
...
+- [ ] (R) CBPP approvers have approved the CBPP status as `implementable`
+- [ ] (R) CBPP summary, motivation and best practice details are appropriately documented
+- [ ] (R) Test plan is in place, giving consideration to CNF Test Suite input
+- [ ] (R) Scoring has been determined
+- [ ]   "Implementation History" section is up-to-date
+- [ ]    Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+## **Summary**
+
+A CNF’s microservice(s) should follow the single concern principle. To help achieve this goal, a CNF’s microservice(s) should have only one process type (or set of parent/child processes) per container. The process(es) in the container should not spawn other process types (e.g. executables) as a way to contribute to the workload but rather should interact with other processes through a microservice API.
```
CsatariGergely, line 29: I would not bound to this to the process structure. It is more important to group the processes of the same functionality to a container, than the process structure. 
Using the process type results in too many unnecessary containers what creates unnecessary overhead and stresses the limits of container orchestration in stead of benefiting from the Linux kernel process scheduling capabilities.

doc/cbpps/0005-single-concern-per-container.md
```diff
@@ -0,0 +1,179 @@
...
+- Reducing downtime risk in deployments and upgrades, increasing the speed of delivery of services
+- Managing service concerns in independent units
+
+
+## **Motivation**
+
+Challenges we think this best practice can help solve:
+
+- Life-cycle management
+    - If you have more than one process type in a container, you will have to manage the lifecycle of the secondary concerns within the container, which effectively means creating an orchestrator within the parent process type, reducing the value of using the Kubernetes orchestration capabilities.
```
CsatariGergely: this is not entirely true. Processes are well managed by the Linus kernel, there is no need for a separate orchestrator even if there are multiple process types in a container.

doc/cbpps/0005-single-concern-per-container.md
```diff
@@ -0,0 +1,179 @@
...
+    - Resource utilization is likely to be less efficient in multi-concerned containers, as the container runtime will look to allocation infrastructure resources for all components rather than individual microservices requirements.
+    - Response time of multi-concerned containers is increased as scaling is required for the combined services rather than the individual services needing scaling.
```
CsatariGergely: On the other hand there is an overhead of the separation of processes to containers.

CsatariGergely (changes requested):

doc/cbpps/0005-single-concern-per-container.md
jeffsaelens: This is slightly subjective. What constitutes "too many" containers? There is definitely a breaking point where the burden of the platform overhead outweighs the advantages gained from the platform itself, but how do we know the default in this scenario will be that we have too many containers? The opposite POV is that we have "too few containers" where big bloated VM like containers are being pushed across the network and when it goes down the failure domain is increased. 

I think your concern is valid, but I think the notion of responsibly thinking about what processes you map to a container and sticking to the UNIX principles here are still valid. There is a happy medium somewhere in here, and striving towards a best practice isn't the same as being beholden to an ironclad rule.

doc/cbpps/0005-single-concern-per-container.md
```diff
@@ -0,0 +1,179 @@
...
+- [ ] (R) Test plan is in place, giving consideration to CNF Test Suite input
+- [ ] (R) Scoring has been determined
+- [ ]   "Implementation History" section is up-to-date
+- [ ]    Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+## **Summary**
+
+A CNF’s microservice(s) should follow the single concern principle. To help achieve this goal, a CNF’s microservice(s) should have only one process type (or set of parent/child processes) per container. The process(es) in the container should not spawn other process types (e.g. executables) as a way to contribute to the workload but rather should interact with other processes through a microservice API.
+
+_“It’s best practice to separate areas of concern by using one service per container. That service may fork into multiple processes (for example, the Apache web server starts multiple worker processes). It’s ok to have multiple processes, but to get the most benefit out of Docker, avoid one container being responsible for multiple aspects of your overall application.”_
```
jeffsaelens:
```suggestion
_“It’s a best practice to separate areas of concern by implementing one service per container. That service may fork into multiple processes (I.E. an Apache web server starts multiple worker processes). It is acceptable to have multiple processes running in a container; however, to get the maximum benefit out of a container runtime environment, avoid one container being responsible for multiple aspects of your overall application.”_
```

doc/cbpps/0005-single-concern-per-container.md
jeffsaelens: Where should we define separation of concerns in your opinion? The kernel is indeed a work horse but when its stated there is no need for separate orchestration, wouldn't we at that point just run bare metal implementations leveraging packages and package managers?

doc/cbpps/0005-single-concern-per-container.md
```diff
@@ -0,0 +1,179 @@
...
+- Managing service concerns in independent units
+
+
+## **Motivation**
+
+Challenges we think this best practice can help solve:
+
+- Life-cycle management
+    - If you have more than one process type in a container, you will have to manage the lifecycle of the secondary concerns within the container, which effectively means creating an orchestrator within the parent process type, reducing the value of using the Kubernetes orchestration capabilities.
+    - Resource utilization is likely to be less efficient in multi-concerned containers, as the container runtime will look to allocation infrastructure resources for all components rather than individual microservices requirements.
```
jeffsaelens:
```suggestion
    - Resource utilization is likely to be less efficient in multi-concerned containers, as the container runtime will look to allocate infrastructure resources for all components rather than individual microservices requirements.
```

doc/cbpps/0005-single-concern-per-container.md
```diff
@@ -0,0 +1,179 @@
...
+    - Response time of multi-concerned containers is increased as scaling is required for the combined services rather than the individual services needing scaling.
+    - Upgrades causing interruption in many services and all their integration at once because none of the components in the system are independent 
+- Security
+    - CNFs with multi-concerned containers have a larger surface area for security attacks and production bugs
```
jeffsaelens:
```suggestion
    - CNFs with multi-concerned containers have a larger surface area for cyberattacks and production bugs
```

jeffsaelens: We might want to add some sub-bullets to this such as expanding on what the attack surface enables attackers to achieve. The ability to move laterally (especially if privilege is established)

doc/cbpps/0005-single-concern-per-container.md
```diff
@@ -0,0 +1,179 @@
...
+- Software Development Cycle
+    - A CNF with multiple maintainers (which can be different groups in an organization) can block each other’s development process when a CNF is tightly coupled
+    - A high degree of test coverage with a complex and tightly coupled CNF is difficult to achieve
+    - Enable "polyglot" software designs, where the decoupled entities are developed in entirely different frameworks and even computer languages
+
+
+### **Goals**
+
+- Life-cycle management
+    - Increase the use of Kubernetes orchestration capabilities instead of using additional container orchestration solutions
```
jeffsaelens: I question this goal. Is our end state K8s adoption or better software practices? I'm tracking the CNCF is our backer here, but I hesitate to specifically state we are pushing a purely K8s centric approach vs just trying to optimize orchestration in a cloud native manner. What if the CNCF or another opensource group provides an outstanding K8s alternative sometime in the near future? 

I think this main goal should be more focused on simplifying (or making more consistent) the approach to orchestration via adopting the principle.

doc/cbpps/0005-single-concern-per-container.md
```diff
@@ -0,0 +1,179 @@
...
+    - Enable "polyglot" software designs, where the decoupled entities are developed in entirely different frameworks and even computer languages
+
+
+### **Goals**
+
+- Life-cycle management
+    - Increase the use of Kubernetes orchestration capabilities instead of using additional container orchestration solutions 
+    - Align with microservice architectural practices for operations and development
+    - Make it easier to scale multiple containers of a CNF to produce an efficient resource utilization and faster response:
+        - Make it easier to reason the best way to scale a CNF based on anticipated load by reducing the complexity of service concerns and allowing focus on each container (or Pods) anticipated needs
```
jeffsaelens: I don't think PODs fits here as a pod is where we want to start to aggregate "concerns". I think it's another construct that allows us to more safely implement the best practice proposed. I also think the notion of a pod "potentially" addresses some of @CsatariGergely 's concerns as it allows us to package multiple containers (E.G. processes in this case) together. 

That said, pods for the most part are a K8s resource which I hesitate to fully hitch my wagon to.

doc/cbpps/0005-single-concern-per-container.md
```diff
@@ -0,0 +1,179 @@
...
+### **Goals**
+
+- Life-cycle management
+    - Increase the use of Kubernetes orchestration capabilities instead of using additional container orchestration solutions 
+    - Align with microservice architectural practices for operations and development
+    - Make it easier to scale multiple containers of a CNF to produce an efficient resource utilization and faster response:
+        - Make it easier to reason the best way to scale a CNF based on anticipated load by reducing the complexity of service concerns and allowing focus on each container (or Pods) anticipated needs 
+        - Support scaling individual processes of the same type for efficient resource utilization and faster response time for changes in service load (increasing/decreasing)
+    - Simplify deployment, reduce risk in upgrades and support easier rollbacks by managing the process types (service concerns) independently and providing coarse-grained dependencies at the container level 
+- Security
```
jeffsaelens, line 70: Increases confidentiality by creating clearly defined boundaries (defense in depth) and allows for "need to know" access to be employed via the interfaces you've previously called out. 
Increases integrity (this is covered by the points you have listed)
Increases availability (also by the points you you've laid out but I think we should tie this in).

I don't think we need to go nuts here, but tying this back to the CIA triad will earn us some browning points with the true security wonks I think.

doc/cbpps/0005-single-concern-per-container.md
```diff
@@ -0,0 +1,179 @@
...
+    - Reducing attack surface area in a CNF’s containers by limiting the number process types and their dependencies such as additional binaries/libraries 
+    - Protect containers from interfering with one another by leveraging the container namespace system
+    - Allow finer control of permissions
+- Observability
+    - Simplifying troubleshooting and readability of log output. It is easier to reason about the log messages that come from one concern or process type than log messages that are interleaved with other concerns. This is even more true in a container that prints all log messages to standard out, such as in a 12-factor cloud native app.
```
jeffsaelens:
```suggestion
    - Simplifying troubleshooting and readability of log output. It is easier to consume log messages and reason about their output when they come from one concern or process type as opposed to when they are interweaved with other concerns. This is even more true in a container that prints all log messages to standard out, such as described in the 12-factor cloud native app.
```

doc/cbpps/0005-single-concern-per-container.md
```diff
@@ -0,0 +1,179 @@
...
+- Best practices for CNFs communication with other processes through microservice APIs
+- Best practices on process supervision and management
+- Reduce risk issues from of home-grown process management systems
+    - Security, Zombie processes, unknown failures
+- Specifying implementation details of each component
+- Specifying how a CNF should be split into individual micro services
+
+## **Proposal**
+
+A CNF with multiple concerns should split services (or process types) for each of its concerns into separate containers. Service dependencies should be handled between containers through well-defined interfaces.
```
jeffsaelens: Should this really be phrased around course/fine grain packaging with containers given as an example? I think this principle applies to VMs and BM workloads as well. We don't shove 5 mission critical monolith applications onto a single server these days because we know bad things happen when that server goes down or gets hacked.

doc/cbpps/0005-single-concern-per-container.md
```diff
@@ -0,0 +1,179 @@
...
+
+## **Proposal**
+
+A CNF with multiple concerns should split services (or process types) for each of its concerns into separate containers. Service dependencies should be handled between containers through well-defined interfaces. 
+
+Pod specs for the CNF should provide scaling and monitoring information for each service running in different containers.
+
+## **Workload Context**
+
+All pod types should implement this best practice.
```
jeffsaelens: confused on what we mean by pod type here. This was mostly container focused and now we are saying the implementation happens at the pod level? (could be I just lack context here)

jeffsaelens (changes requested):
I think this is an excellent read. I put in a few thoughts but overall I think that its in a really good place.

doc/cbpps/0005-single-concern-per-container.md
taylor: @CsatariGergely the main intent is **single concern** per container and process type happens to often align with concerns. See references for reasons for this (which is also a principle -- single concern principle) such as Docker on containers, Intel and Ericsson on flexible and programmable networks supporting automation.

The is a best practice suggested for *most* cases not a hard requirement for all cases. When there are specific cases where the benefits of having a multi-concerned container (with multiple process types) leveraging Linux kernel process scheduling outweigh splitting the concerns into multiple containers then simply do not follow this best practice.  Note why that *specific* case is not followed.

doc/cbpps/0005-single-concern-per-container.md
electrocucaracha: @jeffsaelens this paragraph is a quote, which it was taken as it is. Even when your suggestion improves its readability, it breaks its traceability.

taylor: @jeffsaelens this was a direct quote.  Should we fix the grammar?

doc/cbpps/0005-single-concern-per-container.md
electrocucaracha: So, do you suggest removing the word `Kubernetes` from the paragraph or remove the goal entirely?
```suggestion
    - Increase the use of orchestration capabilities instead of using additional container orchestration solutions 
```

doc/cbpps/0005-single-concern-per-container.md
jeffsaelens: If it's being leveraged as a quote then it should not be changed. I didn't see a citation though, hence the suggestion. I think we either paraphrase and clean it up, or reference who is being quoted.

doc/cbpps/0005-single-concern-per-container.md
taylor: @CsatariGergely this is about dealing with processes dying and restarting, scaling the processes appropriately within the container, etc.

doc/cbpps/0005-single-concern-per-container.md
taylor: @CsatariGergely this can be a suggested addition (use the GitHub suggestion feature in the PR) to the Notes/Caveats section.  Something like:

> For some applications and use cases there may be an increased overhead as a result separation of processes into different containers

doc/cbpps/0005-single-concern-per-container.md
taylor: @jeffsaelens please change it to cyberattack without the hyphen.

doc/cbpps/0005-single-concern-per-container.md
taylor: @jeffsaelens Maybe it should say leverage the platforms orchestration capabilities?

doc/cbpps/0005-single-concern-per-container.md
taylor: @jeffsaelens I think "(or Pods)" can be removed an the goal is still good enough for the intent of this best practice.  Please make a suggest edit.

doc/cbpps/0005-single-concern-per-container.md
taylor: @jeffsaelens I agree.  Suggest edit?  :)

doc/cbpps/0005-single-concern-per-container.md
jeffsaelens:
```suggestion
    - Simplify and consolidate the use of orchestration capabilities instead of adding additional container orchestration solutions 
```

jeffsaelens: I think just dropping K8s from the first makes it mildly confusing. I don't know if the suggestion above is really it either, but I took a stab at it.

doc/cbpps/0005-single-concern-per-container.md
taylor: This is based on what we have used for other best practices.  Pod type was part of the context.  I like this definition

> Pod is a set of Linux namespaces, cgroups, and potentially other facets of isolation - the same things that isolate a [container](https://kubernetes.io/docs/concepts/containers/). 

from https://kubernetes.io/docs/concepts/workloads/pods/#what-is-a-pod

For some best practices we thought they should not apply to K8s core-system Pod types. 

In any case we could have other things here such as control and user plane applications if desired.

doc/cbpps/0005-single-concern-per-container.md
jeffsaelens:
```suggestion
        - Make it easier to reason the best way to scale a CNF based on anticipated load by reducing the complexity of service concerns and allowing focus on each container's anticipated needs 
```

doc/cbpps/0005-single-concern-per-container.md
jeffsaelens: ignore last, I'm looking at the chopped up reviewers view and missed the reference and formatting. When I looked at the finished view, its very obvious this is a quote. Disregard this.

doc/cbpps/0005-single-concern-per-container.md
jeffsaelens: If we stay K8s centric I think it makes sense. If we are more generic, I'm iffy.

jeffsaelens (commented):
I'll work on a security/triad edit here shortly.

doc/cbpps/0005-single-concern-per-container.md
taylor: Remember this is a *cloud native* best practice for applications.  This is not trying to be the best practice for all things.  It may not be a practice that fits every situation, but that does not change if it is a *cloud native* best practice or not.

doc/cbpps/0005-single-concern-per-container.md
taylor: These are currently in the vast majority of cases applied to K8s environments.  We can shift to something more generic later as needed. All of these practices may be modified as needed going forward.

doc/cbpps/0005-single-concern-per-container.md
```diff
@@ -0,0 +1,177 @@
...
```
taylor:
```suggestion
All Kubernetes Pod types should implement this best practice.  This best practice applies to control and user plane applications.
```

doc/cbpps/0005-single-concern-per-container.md
taylor:
```suggestion
    - Security vulnerabilities in one process type may affect all other processes in the same container
```

doc/cbpps/0005-single-concern-per-container.md
taylor: Suggested update added

doc/cbpps/0005-single-concern-per-container.md
electrocucaracha: Containers, among other things, provide isolation (namespaces) and resources control (cgroups). Having a group of different process in the same container provides the same limit resource, e.g. CPU time-share. A `sshd` process type may require much higher CPU share than the rest of processes. In the other hand, inter-containers communication can also be accomplished by using [shared files on the disk](https://kubernetes.io/docs/tasks/access-application-cluster/communicate-containers-same-pod-shared-volume/#discussion). BTW, containers in the same pod [share the same IPC namespace](https://www.mirantis.com/blog/multi-container-pods-and-container-communication-in-kubernetes/).

doc/cbpps/0005-single-concern-per-container.md
taylor: The [previous reminder](https://github.com/cncf/cnf-wg/pull/267#discussion_r1324803547), about this being a cloud native best practice, applies here as well.

taylor:
> I think this is not a best practice for CNF-s. Restricting to single concern or process type per container results in too many API calls in the architecture what gives a too big overhead to the latency sensitive CNF-s.

This applies to the majority of applications including networking-focused applications.  Specifically, it is a **cloud native** best practice for the majority of networking applications.

For latency-sensitive applications, specifically the individual services (and their processes) within the application that are latency sensitive, when this best practice's benefits are outweighed by the other practices meeting latency requirements an exception in following this best practice should be made for that specific application's services (and their processes).  

This information is a good candidate for the NOTES and caveats section.

doc/cbpps/0005-single-concern-per-container.md
```diff
@@ -0,0 +1,177 @@
...
+1. https://docs.docker.com/config/containers/multi-service_container/
+1. https://en.wikipedia.org/wiki/Single-responsibility_principle
+1. https://developers.redhat.com/articles/2022/01/11/5-design-principles-microservices
+1. https://www.ericsson.com/en/future-technologies/architecture/network-capabilities
+_“As a step towards a fully autonomous network and achieving an intent-based management of a network, its architecture must be prepared by raising the level of abstraction in management with e.g., strong separation of concerns.”_
+1. https://github.com/cncf/cnf-testsuite/blob/main/docs/LIST_OF_TESTS.md#single-process-type-in-one-container
+    - RATIONALE - https://github.com/cncf/cnf-testsuite/blob/main/RATIONALE.md#to-check-if-the-cnf-has-multiple-process-types-within-one-container-single_process_type
+1. https://www.tutorialworks.com/containers-single-or-multiple-processes/
+1. https://www.intel.com/content/dam/www/public/us/en/documents/white-papers/containers-and-cloud-native-functions-white-paper.pdf
+1. https://techcommunity.microsoft.com/t5/azure-for-operators-blog/what-is-the-5g-session-management-function-smf/ba-p/3693852
```
electrocucaracha: I found another interesting article from Mirantis
```suggestion
1. https://techcommunity.microsoft.com/t5/azure-for-operators-blog/what-is-the-5g-session-management-function-smf/ba-p/3693852
1. https://www.mirantis.com/blog/multi-container-pods-and-container-communication-in-kubernetes/
```

doc/cbpps/0005-single-concern-per-container.md
jeffsaelens: I made a PR against this branch with an attempt.

taylor: https://github.com/cncf/cnf-wg/pull/269

Smitholi67 (approved):
lgtm

doc/cbpps/0005-single-concern-per-container.md
taylor: merged

doc/cbpps/0005-single-concern-per-container.md
```diff
@@ -0,0 +1,175 @@
...
+  - Simplify and consolidate the use of orchestration capabilities instead of adding additional container orchestration solutions
+  - Align with microservice architectural practices for operations and development
+  - Make it easier to scale multiple containers of a CNF to produce an efficient resource utilization and faster response:
+    - Make it easier to reason the best way to scale a CNF based on anticipated load by reducing the complexity of service concerns and allowing focus on each container's anticipated needs
+    - Support scaling individual processes of the same type for efficient resource utilization and faster response time for changes in service load (increasing/decreasing)
+  - Simplify deployment, reduce risk in upgrades and support easier rollbacks by managing the process types (service concerns) independently and providing coarse-grained dependencies at the container level
+- Security
+  - Increase confidentiality by creating clearly defined boundaries between software components
+    - Reduces the attack surface presented in a CNF’s containers
+    - Limit the number of process types and their dependencies such as additional binaries/libraries
```
lixuna:
```suggestion
    - Limits the number of process types and their dependencies such as additional binaries/libraries  
```

doc/cbpps/0005-single-concern-per-container.md
```diff
@@ -0,0 +1,175 @@
...
+  - Align with microservice architectural practices for operations and development
+  - Make it easier to scale multiple containers of a CNF to produce an efficient resource utilization and faster response:
+    - Make it easier to reason the best way to scale a CNF based on anticipated load by reducing the complexity of service concerns and allowing focus on each container's anticipated needs
+    - Support scaling individual processes of the same type for efficient resource utilization and faster response time for changes in service load (increasing/decreasing)
+  - Simplify deployment, reduce risk in upgrades and support easier rollbacks by managing the process types (service concerns) independently and providing coarse-grained dependencies at the container level
+- Security
+  - Increase confidentiality by creating clearly defined boundaries between software components
+    - Reduces the attack surface presented in a CNF’s containers
+    - Limit the number of process types and their dependencies such as additional binaries/libraries  
+    - Prevent unnecessary access to data via shared filesystems
```
lixuna:
```suggestion
    - Prevents unnecessary access to data via shared filesystems
```

doc/cbpps/0005-single-concern-per-container.md
```diff
@@ -0,0 +1,175 @@
...
+    - Implement finer grained attribute-based controls between processes and systems
```
lixuna:
```suggestion
    - Allows for finer grained attribute-based controls to be implemented between processes and systems
```

doc/cbpps/0005-single-concern-per-container.md
```diff
@@ -0,0 +1,175 @@
...
+  - Increase integrity by restricting a processes ability to perform CRUD operations in a shared container environment
+    - Protect containers from interfering with one another by leveraging the container namespace system and cgroup implementations
```
lixuna:
```suggestion
    - Protects containers from interfering with one another by leveraging the container namespace system and cgroup implementations
```

doc/cbpps/0005-single-concern-per-container.md
```diff
@@ -0,0 +1,175 @@
...
+  - Increase availability by reducing the size of failure domain
+    - Allows for granular control of separation of concerns
+    - Remove the risk of multiple critical processes terminating simultaneously
```
lixuna:
```suggestion
    - Removes the risk of multiple critical processes terminating simultaneously
```

doc/cbpps/0005-single-concern-per-container.md
```diff
@@ -0,0 +1,175 @@
...
+    - Define access methods via established interfaces
```
lixuna:
```suggestion
    - Defined access methods via established interfaces
```

lixuna (changes requested):

doc/cbpps/0005-single-concern-per-container.md
lixuna:
```suggestion
    - Defines access methods via established interfaces
```

taylor:
```suggestion
    - Defines access methods via established interfaces
```

lixuna (approved):
lgtm

jeffsaelens (approved):
Updates are solid. LGTM

agentpoyo (approved):
LGTM :+1:

electrocucaracha (approved):
/lgtm

doc/cbpps/0005-single-concern-per-container.md
CsatariGergely: Thanks for the reactions. 
@jeffsaelens I agree that this is subjective and we need to find a medium. I think connecting directly to process types is a bit to rigid practice. I see now that this we corrected. 

@taylor Thanks for the updates. I think we should put more emphasis on having the same service or functionality in a container than process types.

nickolaev (approved):

sparso (approved):
LGTM

taylor:
Thanks for everyone's reviews, suggest edits, and comments.  There are no remaining items left and we have more than met the minimum approvals for merging so I will squash and merge.

rich-l (approved):

ildikov (approved):

